PR #9, opened 2016-08-01
newsletter: added variable config to decide whether to send welcome m…

ghost:
…ails to

  newsletter subscribers


torotil:
The changes for the %confirm_url description are part of this PR since they went into the same branch.

torotil:
That's solved in another way by now anyway.

PR #10, opened 2016-08-05
wizard Email: added help text for the confirm_url token for confirmation

ghost:
request emails


torotil:
That's outdated now that we use webform4.

PR #11, opened 2016-11-08
Embedded thankyou page

alex-mo:
If a form is embedded via an iframe and `webform_ajax` loads the thank you page, then `X-Frame-Options` should be set.

### DIFF
--- a/campaignion_webform_widget/campaignion_webform_widget.module
+++ b/campaignion_webform_widget/campaignion_webform_widget.module
@@ -32,6 +32,22 @@ function campaignion_webform_widget_form_webform_client_form_alter(&$form, &$for
 }
 
 /**
+ * Implements hook_node_view_alter().
+ */
+function campaignion_webform_widget_node_view_alter(&$build) {
+  $params = drupal_get_query_parameters();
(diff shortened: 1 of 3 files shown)


alex-mo:
Missing pieces:
- thankyou page needs to know it's embedded
- any links in the thankyou page should be opened in `target=_blank` (the embedded journey is done with the thankyou page)

alex-mo:
the thankyou page now knows if it embedded, if a (checkbox) field named `field_pass_embed` is created on the action and is toggled on.

alex-mo:
you will need to:
- allow `body` field to be displayed as CCK block
- create a context for the TY page, dependending on `current_webform_embedded` context which displays the `body` in the region `widget`
- check if the template `page--current-webform-embedded.tpl.php` is up-to-date (for the title to be displayed)

torotil:
Thanks for the work and thoughts on this.

As discussed I don't intend to merge this unless we have a clear overall picture about how our widget is intended to work in the first place.

alex-mo:
one more todo:
* multi step forms with webform_ajax currently break these patches

PR #13, opened 2017-01-09
Mailchimp v3

torotil:
API version upgrade for the campaignion mailchimp integration.


torotil:
The tests still use the old version of little_helpers. That's why they fail.

PR #17, opened 2017-02-14
[cleverreach] Only set registered date for new subscriptions.

torotil:
Stop campaignion from overriding the registered date of existing subscribers.


torotil:
This can go live as a quck-fix but it should have tests. The class can easily be made unit-testable.

PR #18, opened 2017-02-21
Move over validation from profile.

torotil:
This is should go with moreonion/campaignion_starterkit#12.


torotil:
Needs work:
* The field name should be a parameter of the action-type.
* Check that the field type is also defined in the campaignion module.

PR #24, opened 2017-03-09
Email token expiry

torotil:
Depends on moreonion/webform_confirm_email#1 .

### DIFF
--- a/campaignion_wizard/campaignion_wizard.install
+++ b/campaignion_wizard/campaignion_wizard.install
@@ -27,6 +27,24 @@ function campaignion_wizard_schema() {
 }
 
 /**
+ * Implements hook_update_dependencies().
+ */
+function campaignion_wizard_update_dependencies() {
+  $deps['campaignion_wizard'][3]['webform_confirm_email'] = 7205;
+  return $deps;
+}
+
+/**
+ * Update the request_lifetime of all confirmation emails.
+ */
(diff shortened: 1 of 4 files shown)


campaignion_wizard/campaignion_wizard.install
```diff
@@ -27,6 +27,25 @@ function campaignion_wizard_schema() {
 }
 
 /**
+ * Update the request_lifetime of all confirmation emails.
+ */
+function campaignion_wizard_update_3() {
+  $transaction = db_transaction();
```
torotil: There is no need to use transactions for a single update statement.

campaignion_wizard/src/Email.php
```diff
@@ -214,21 +224,137 @@ protected function saveEmail($email_type, &$form, &$form_state) {
...
-      ->fields(array('email_type' => $email_type))
+      ->fields($fields)
       ->execute();
   }
 
+  /**
+   * Inserts a field labeled "Advanced Settings" which contains fields for
+   * configuring the request lifetime settings of an email.
+   */
+  protected function insertLifetimeFields($email, &$email_form) {
```
torotil: We should try to leverage whatever `webform_confirm_email` provides in terms of form elements here.

torotil (commented):
Needs work / a slightly different approach.

torotil:
Completely redid this.

PR #26, opened 2017-03-23
Add SourceInterface::hasKey() and implement it in all sources.

torotil:
Motivation is to be able to distinguish between sources providing an empty value and sources providing no value.


torotil:
There should be an implementation of SourceInterface for the contact (for campaignion_newsletters).

PR #33, opened 2017-04-27
[e2t] fix IE11 on Win7 error: build js with patched version of Vue 1.0

mayakokits:


torotil:
@katzmo and me have tried to figure out what the actual change is and we could find a commit. Is there a pull request for vue.js for this change? We think there should be: We don't want to maintain a vue fork. ;)

torotil:
After discussing: Let’s wait for our port to vue 2 which should happen in the next month anyway.

PR #50, opened 2017-09-21
Behavior sort hack webform_scroll is executed last.

torotil:


torotil:
The solution is a bit hacky but it seems to work for all browsers and is good enough for now.

katzmo:
Since Drupal behaviors are also called on `$(document).ready()` (`$()`), can you be sure this will always run last?

torotil:
Usually behaviors are *defined* directly in the JavaScript file (in a `function($) {}(jQuery)`) wrapper. This patch postpones the definition for this feature until `$(document).ready()`. There is a slight risk, that this behavior won’t be executed as behaviors (I think) also *executed* when the page is first built. But for this specific behavior that’s not an issue. It wouldn’t do anything anyway.

PR #55, opened 2017-10-04
Display warning instead of template form if there are submissions.

m-strasser:
Form template selection is now contained in a fieldset at the bottom of the page. If the form has already been submitted, the select and submit button get replaced by a warning text.

### DIFF
--- a/campaignion_wizard/src/WebformStep.php
+++ b/campaignion_wizard/src/WebformStep.php
@@ -31,10 +31,39 @@ class WebformStep extends WizardStep {
     $build['#attached']['js'][] = drupal_get_path('module', 'campaignion_wizard') . '/js/form-builder-submit.js';
     $build['#attached']['library'][] = array('system', 'ui.datepicker');
 
+    // Remove #title from custom submit buttons fieldset.
+    if (module_exists('webform_custom_buttons')) {
+      $form['submit_buttons']['#title'] = '';
+    }
+
     // Build form for webform_template select box.
     if (module_exists('campaignion_action_template')) {
-      $build[] = drupal_get_form('campaignion_action_template_selector_form', $this->wizard->node)
-        + array('#weight' => -99);
+      $node = $this->wizard->node;
+      // Check if there are existing form submissions.
+      $sql = "SELECT 1 FROM {webform_submissions} WHERE nid=:nid LIMIT 1";
+      $has_submissions = db_query($sql, [':nid' => $node->nid])->fetchField();
+      $webform_template_form = drupal_get_form('campaignion_action_template_selector_form', $node);
+      $webform_template_form['#weight'] = 1;
+
+      if ($has_submissions) {
+        $warning_msg = t('Applying form templates is not possible on forms that have already captured data. ' .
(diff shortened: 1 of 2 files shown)


campaignion_wizard/src/WebformStep.php
```diff
@@ -33,8 +33,42 @@ public function pageCallback() {
 
     // Build form for webform_template select box.
     if (module_exists('campaignion_action_template')) {
+      // Check if there are existing form submissions.
+      $submissions_exist = FALSE;
+      $sql = <<<SQL
+SELECT count(*) AS submissions
+FROM {webform_submissions}
+WHERE nid=:nid
+SQL;
```
torotil: I think something like this would be better (no need to count them all).

```php
$has_submissions = (bool) db_select('webform_submissions', 's')
  ->fields('s', ['sid'])
  ->condition('nid', $nid)
  ->range(0, 1)
  ->execute()
  ->fetchField();
```

or without the using as much of the db-abstraction layer:

```php
$q = 'SELECT 1 FROM {webform_submissions} WHERE nid=:nid LIMIT 1';
$has_submissions = db_query($q, [':nid' => $nid])->fetchField();
```

campaignion_wizard/src/WebformStep.php
```diff
@@ -33,8 +33,42 @@ public function pageCallback() {
...
+      $submissions = db_query($sql, array(':nid' => $this->wizard->node->nid));
+      foreach ($submissions as $s) {
+        if ($s->submissions > 0) {
+          $submissions_exist = TRUE;
+        }
+      }
+
       $build[] = drupal_get_form('campaignion_action_template_selector_form', $this->wizard->node)
```
torotil: I think it would be better to generate the form into a temporary variable first, then modify it, and then assign it to `$build[]`. This way we would avoid to assign values to a "magic" index.

campaignion_wizard/src/WebformStep.php
```diff
@@ -33,8 +33,42 @@ public function pageCallback() {
...
+        $build[1]['webform_template']['webform_template_warning'] = array(
+          '#markup' => '<div class="action-template-warning">' .
+          t('Applying form templates is not possible on forms that have already captured data. If you want to apply a form template you have to <a href="@delete-submissions">delete all form data first</a>.',
+            array(
+              '@delete-submissions' => url(
+                'node/' . $this->wizard->node->nid . '/webform-results/clear',
+                array(
+                  'query' => array(
+                    'destination' => 'node/' . $this->wizard->node->nid . '/wizard/form'
+                  ),
```
torotil: Please use the new style array syntax for new code.

Also splitting this into multiple statements would avoid such deeply nested arrays. Also you can perhaps use `$nid` here as you do further up.

```php
$o['query']['destination'] = url("node/$nid/wizard/form");
$url = url("node/$nid/webform-results/clear", $o);
$warning = t("…", ['@delete-submissions' => $url]);
```

torotil (changes requested):

torotil (approved):

PR #71, opened 2017-11-17
Improve PayPal redirect message.

torotil:


katzmo (approved):
Sounds nicer, though I'd prefer "donation" instead of "payment"…

torotil:
This could be used in a non-donation context though — even in campaignion installations.

PR #73, opened 2017-11-23
Add missing field_base:supporter_tags.

torotil:
The field is used in the campaignion_supporter_tags module so the field_base should be declared there. This was simply missed by the previous pull requests.


katzmo (approved):
Approve of adding a name, too!

PR #76, opened 2017-12-13
[campaignion_supporter_tags] taxonomy names are trimmed

alex-mo:
- taxonomy_term_save() enforces trimmed names


torotil:
The comment doesn’t seem to be coding-standards compliant (phpcs-drupal).

torotil:
I’ve modified the commit a bit and also changed the case to lower-case for comparisons. This way we don’t get duplicate tags for this reason either.

alex-mo (commented):
👍

PR #77, opened 2017-12-19
3-state optins: subscribe / unsubscribe / no change

torotil:
* Changes how values are stored:
  * 'subscribed': checkbox checked OR radio yes [unchanged]
  * 'unsubscribed': radio no [NEW]
  * '': checkbox not checked OR radio not selected [unchanged]
* When the value is 'unsubscribed' the email address is unsubscribed
  from all known lists.

This will be made configurable in the future, but it’s enough for some
special cases for the while.

### DIFF
--- a/campaignion_newsletters/campaignion_newsletters.module
+++ b/campaignion_newsletters/campaignion_newsletters.module
@@ -96,10 +96,11 @@ function campaignion_newsletters_webform_submission_confirmed(Submission $submis
     return;
   }
   foreach ($s->webform->componentsByType('newsletter') as $component) {
-    if ($s->valueByCid($component['cid'])) {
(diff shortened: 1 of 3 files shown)


campaignion_newsletters/campaignion_newsletters.module
```diff
@@ -110,6 +111,11 @@ function campaignion_newsletters_webform_submission_confirmed(Submission $submis
         }
       }
     }
+    elseif ($value == 'unsubcribed') {
+      foreach (Subscription::byEmail($email) as $s) {
```
alex-mo: `$s` needs to be `$subscription`

alex-mo (changes requested):
variable name mismatch

alex-mo (approved):

PR #80, opened 2018-01-12
Personalized redirect

torotil:
This includes:

* A vue app to deal with the configuration.
* A configuration storage API in campaignion_action.
* The integration in the wizard ThankyouStep.
* The actual implementation of the redirect behaviour in campaignion_action.


campaignion_action/src/ActionBase.php
```diff
@@ -121,11 +121,11 @@ public function redirect(Submission $submission, $delta) {
       case 'redirect':
         $redirects = Redirect::byNid($this->node->nid, $delta);
         foreach ($redirects as $r) {
-        if ($r->checkFilters($submission)) {
-          return $r->normalized();
+          if ($r->checkFilters($submission)) {
+            return $r->normalized();
+          }
         }
         break;
```
alex-mo: dangling `break`. can be dropped

alex-mo:
* +1 for the usage of constants
* +1 for meaningful comments
* `break` unnecessary? (see inline comment)
* probably also test for the non-check optin return `FALSE` as this should not happen
* is there a specific reasoning making the functions `static`? (the commit with the `#process` fn)

alex-mo (changes requested):

campaignion_action/src/ActionBase.php
torotil: I’d vote to keep. Think of someone editing the code in the future in a way that doesn’t lead to a return in any case …

torotil:
Yes, there is a reason to use static functions here: The `#process` property is stored in the form cache (the whole `$form` is cached there). So adding a `[$this, 'addRedirectSettigs']` as callback would put the whole form object there. This means:

1. The step object needs to be serialisable.
2. We add lots of redundant data to the form cache.
3. We have another (older) copy of the step object in there in later requests. Sometimes the "new" one is used, sometimes the old one is used. This adds loads of complexity.

torotil:
Added the additional test for the negated opt-in filter.

campaignion_action/src/ActionBase.php
alex-mo: ok with me...

alex-mo (approved):

PR #81, opened 2018-01-27
Address importer: Support for single-value address fields.

torotil:


src/CRM/Import/Field/Address.php
```diff
@@ -52,41 +52,101 @@ public function import(SourceInterface $source, \EntityMetadataWrapper $entity)
...
+   * @param array $a1
+   *   The first address.
+   * @param array $a2
+   *   The scond address.
+   *
+   * @return bool
+   *   TRUE when the first address contains no non-NULL values that differ from
+   *   the second address.
+   */
+  protected function addressIsSubset($a1, $a2) {
```
alex-mo: better called `addressIsMergeable()` as this better catches the intention

alex-mo (changes requested):

alex-mo (approved):

PR #84, opened 2018-02-15
Feature publish content permission

m-strasser:


campaignion_wizard/src/ConfirmStep.php
```diff
@@ -52,15 +52,22 @@ protected function generateStatusMessages() {
...
    */
-  protected function generateButtons($form) {
+  protected function generateButtons($form, &$form_state) {
     unset($form['buttons']['previous']);
     $buttons = [
       '#weight' => 1000,
       '#attributes' => ['class' => ['form-submit']],
     ] + $form['buttons'];
 
+    $node = $form_state['oowizard']->node;
```
torotil: You can simply use `$this->wizard->node`.

campaignion_wizard/src/ConfirmStep.php
```diff
@@ -52,15 +52,22 @@ protected function generateStatusMessages() {
...
     $can_publish = !module_exists('change_publishing_status_permission')
-      || user_access('change publishing status');
+      || user_access('change publishing status')
+      || (!user_access('change publishing status')
```
torotil: The checks for `!user_access(…)` are redundant.

Maybe we can improve readability when we write it like:

```php
if (module_exists('…')) {
  $buttons['return']['#access'] = user_access('…') || $node->status;  // logically equivalent to $can_publish
  $buttons['draft']['#access'] = user_access('…') || !$node->status;  // logically equivalent to $can_draft
}
```

campaignion_wizard/src/ConfirmStep.php
```diff
@@ -90,7 +98,7 @@ public function stepForm($form, &$form_state) {
       '#type' => 'container',
     ] + $this->generateStatusMessages();
 
-    $form['confirm_container']['buttons'] = $this->generateButtons($form);
+    $form['confirm_container']['buttons'] = $this->generateButtons($form, $form_state);
```
torotil: No need for that if you use `$this->wizard->node`.

torotil (changes requested):
Logic looks sound, but it can be simplified a bit. Thanks!

torotil (approved):

torotil:
Cool! I’ll probably squash the commits a bit before merging, but it’s good to go.

PR #85, opened 2018-02-16
Check if user has 'update' permission for nodes

m-strasser:
Prevents users from circumventing access controls for (un)publishing nodes via batch jobs.

### DIFF
--- a/campaignion_manage/src/BulkOp/ContentPublish.php
+++ b/campaignion_manage/src/BulkOp/ContentPublish.php
@@ -19,12 +19,21 @@ STR;
     $element['warn']['message']['#markup'] = t($message, array('!count' => '<span class="bulkop-count"></span>'));
   }
   public function apply($nids, $values) {
+    $messages = [];
     $nodes = node_load_multiple($nids);
+
     foreach ($nodes as $node) {
       if (!$node->status) {
-        node_publish_action($node);
-        node_save($node);
+        if (node_access('update', $node)) {
+          node_publish_action($node);
+          node_save($node);
+        }
+        else {
+          array_push($messages, t("Could not publish '!node' due to lacking permissions.", ['!node' => $node->title]));
+        }
       }
     }
+
+    return $messages;
   }
 }
--- a/campaignion_manage/src/BulkOp/ContentUnpublish.php
+++ b/campaignion_manage/src/BulkOp/ContentUnpublish.php
@@ -21,12 +21,21 @@ STR;
     $element['warn']['message']['#markup'] = t($message, array('!count' => '<span class="bulkop-count"></span>'));
   }
   public function apply($nids, $values) {
+    $messages = [];
     $nodes = node_load_multiple($nids);
+
     foreach ($nodes as $node) {
       if ($node->status) {
-        node_unpublish_action($node);
-        node_save($node);
+        if (node_access('update', $node)) {
+          node_unpublish_action($node);
+          node_save($node);
+        }
+        else {
+          array_push($messages, t("Could not unpublish '!node' due to lacking permissions.", ['!node' => $node->title]));
+        }
       }
     }
+
+    return $messages;
   }
 }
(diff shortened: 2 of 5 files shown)


torotil:
That perhaps needs changes in ContentListing.php too to only show the links when the user has the permission.

Also we might want to report the fact that the selected bulk operation was not applied to some of the nodes because of missing permissions.

campaignion_manage/src/ContentListing.php
```diff
@@ -127,11 +127,14 @@ protected function nodeLinks($node) {
       }
     }
     foreach (array($edit_path_part => t('Edit'), 'translate' => t('Translate'), 'view' => t('View page'), 'delete' => t('Delete')) as $path => $title) {
-      $links[$path] = array(
-        'href' => "node/{$node->nid}/$path",
-        'title' => $title,
-        'query' => array('destination' => 'admin/manage/content_and_actions')
-      );
+      if (node_access($path, $node)
+         || ($path === 'wizard' && node_access('update', $node))) {
```
torotil: Could we write this more like this?

```php
$action = $path == 'wizard' ? 'update' : $path;
if (node_access($action, $node)) {
```

I think it’s more readable this way. Also it avoids calling `node_access('wizard', $node)`.

campaignion_manage/tests/UnPublishTest.php
```diff
@@ -0,0 +1,49 @@
...
+   * Checks if anonymous users are disallowed from (un)publishing.
+   */
+  public function testAnonymousUser() {
+    $publish_job = new ContentPublish();
+    $msgs = $publish_job->apply([$this->node->nid], []);
+    $this->assertNotEmpty($msgs);
+
+    $unpublish_job = new ContentUnpublish();
+    $msgs = $unpublish_job->apply([$this->node->nid], []);
+    $this->assertEmpty($msgs);
```
torotil: Could you add a test that shows that unpublishing a published node doesn’t work either

campaignion_manage/src/BulkOp/ContentPublish.php
```diff
@@ -19,12 +19,22 @@ public function formElement(&$element, &$form_state) {
     $element['warn']['message']['#markup'] = t($message, array('!count' => '<span class="bulkop-count"></span>'));
   }
   public function apply($nids, $values) {
+    $messages = [];
+    $error_message = "Could not publish \"!node\" due to lacking permissions.";
```
torotil: Please use single quotes in this case to avoid backslashes in the string. Also we should pass this through `t()` directly otherwise .pot-extractors have difficulties picking up the string.

campaignion_manage/src/BulkOp/ContentUnpublish.php
```diff
@@ -21,12 +21,22 @@ public function formElement(&$element, &$form_state) {
     $element['warn']['message']['#markup'] = t($message, array('!count' => '<span class="bulkop-count"></span>'));
   }
   public function apply($nids, $values) {
+    $messages = [];
+    $error_message = "Could not unpublish \"!node\" due to lacking permissions.";
```
torotil: Same as the other error message.

torotil (changes requested):
A few small issues otherwise fine.

torotil (approved):

torotil:
Hm, the whole issue might be bogus. The whole manage page is not accessible unless you have the `administer nodes` permission which grants you all permissions on all nodes:

https://github.com/moreonion/campaignion/blob/0ee7412be800eb9ca8cf7031bca5b4aa22e4a8ac/campaignion_manage/campaignion_manage.module#L5-L36

m-strasser:
The code simply checks if the user has `node_access('update'/'view'/etc., $node)` permissions for a given node and hides buttons accordingly or permits bulk actions if permissions are lacking. The `editor` role with default permissions in `pristine-camp` is an example of a user with that permission set. If `administer nodes` should grant access to CRUD operations on all content types the permissions should be set correctly and the bug lies there. I see no reason not the implement access checks on node level, otherwise we're always open to access level bypasses.

torotil:
Right. The additional checks don’t hurt and the additional tests are good too.

PR #88, opened 2018-02-21
add opt-in statement to newsletter subscription

alex-mo:
- provide a field on the component to set a statement for internal
  audit trail (will not be shown to users)
- save the text along the newsletter_subscription activity
- the activity is saved on confirmation (not on creation)

### DIFF
--- a/campaignion_newsletters/src/Subscription.php
+++ b/campaignion_newsletters/src/Subscription.php
@@ -13,6 +13,7 @@ class Subscription extends \Drupal\little_helpers\DB\Model {
   public $source = NULL;
   public $needs_opt_in = FALSE;
   public $send_welcome = FALSE;
+  public $optin_statement = '';
   public $optin_info = NULL;
 
   public static $lists = array();
--- a/campaignion_newsletters/src/Subscription.php
+++ b/campaignion_newsletters/src/Subscription.php
@@ -2,7 +2,12 @@
 
 namespace Drupal\campaignion_newsletters;
 
-class Subscription extends \Drupal\little_helpers\DB\Model {
+use Drupal\little_helpers\DB\Model;
+
+/**
+ * Subscription model.
+ */
+class Subscription extends Model {
   public $list_id;
   public $email;
   public $fingerprint = '';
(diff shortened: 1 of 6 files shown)


torotil (changes requested):
Looks good in general. Could you try to also meet these criteria?

* Fix coding-style issues in all non-features files touched by the change (in a separate commit).
* Add tests (for the changes) so that the coverage over-all coverage doesn’t decrease.

I know this sets the bar higher for new code than it was for the old, but then again that’s the point isn’t it?

campaignion_newsletters/src/Subscription.php
```diff
@@ -23,6 +28,14 @@ class Subscription extends \Drupal\little_helpers\DB\Model {
   protected static $values = array('fingerprint', 'updated', 'last_sync');
   protected static $serial = FALSE;
 
+  /**
+   * Construct a new instance from data found in the DB.
```
torotil: Mh that should be more like "Get an instance form DB or create a new one." or something like that.

campaignion_newsletters/src/Subscription.php
```diff
@@ -31,18 +44,36 @@ public static function byData($list_id, $email) {
...
+   */
   public static function fromData($list_id, $email) {
     return new static(array(
       'list_id' => $list_id,
       'email' => $email,
     ), TRUE);
   }
 
+  /**
+   * Construct a new instance from email addresses found in the DB.
```
torotil: That should be "Load all subscriptions for one email address." or something similar.

campaignion_newsletters/src/Subscription.php
```diff
@@ -55,6 +86,9 @@ public static function byEmail($email) {
     return $subscriptions;
   }
 
+  /**
+   * Get the newsletter list instance.
```
torotil, line 90: This needs a @return docs.

torotil (commented):
Some of the method summaries are still a bit imprecise, but it definitely looks much better now.

campaignion_newsletters/src/Subscription.php
alex-mo: fine with me

campaignion_newsletters/src/Subscription.php
alex-mo: `phpcs` is silent about that...

torotil (approved):

torotil:
Great, thanks!

PR #90, opened 2018-02-28
Add proper conditional support for campaignion_newsletters.

torotil:


torotil:
@katzmo I’ve added a more elaborate test for whether we need the JavaScript. It’s now added whenever the webform has a conditional based on this conditional_type. Ideally it would still check whether this actually means it has a same-page (aka JS) conditional.

I’ve also taken a look when webform includes it’s own conditional callbacks: It does it whenever a webform is rendered. So there is no on-demand loading at all.

PR #95, opened 2018-03-14
Make mp_fields more robust and tested.

torotil:


katzmo:
Great, that all looks tremendously helpful!
Are you sure that `catch (ConfigError $e) { return NULL; }` has no unwanted side-effects? Maybe we still want to log that the API wasn't available?

katzmo (approved):

PR #100, opened 2018-04-12
Opt in

torotil:
Provide generic opt-in functionality (especially for phone and post).

* Create a new webform component that has similar features as the newsletter component, but without the newsletter list integration.
* Record opt-ins in activities.
* Add contact fields in campaignion starterkit.
* …


campaignion_opt_in/tests/OptInRecordTest.php
```diff
@@ -0,0 +1,84 @@
...
+  public function tearDown() {
+    $this->activity->delete();
+    db_delete('campaignion_opt_in')->execute();
+    parent::tearDown();
+  }
+
+  /**
+   * Test a webform submission with an opt_in and a newsletter component.
+   */
+  public function testTwoOptIuns() {
```
katzmo: typo

campaignion_opt_in/campaignion_opt_in.module
```diff
@@ -47,9 +48,9 @@ function campaignion_opt_in_webform_conditional_form($node) {
       $element = [
         '#type' => 'select',
         '#options' => [
-          'yes' => t('Opt-in'),
-          '' => t('No change'),
-          'no' => t('Opt-out'),
+          Values::OPT_IN => t('Opt-in'),
+          Values::NO_CHANGE => t('No change'),
+          Values::OPT_OUT => t('Opt-out'),
```
katzmo: conditionals need not selected option too.

katzmo (changes requested):

campaignion_opt_in/campaignion_opt_in.module
torotil: Hm, thinking about this again and I don’t think it’s actually needed after all.

campaignion_opt_in/campaignion_opt_in.conditionals.js
```diff
@@ -10,8 +10,10 @@
   Drupal.webform.conditionalOperatorOptInEqual = function (element, existingValue, ruleValue) {
     var checkbox = element.querySelector('.form-type-checkbox input');
     if (checkbox) {
-      return checkbox.checked ? ruleValue === 'checkbox:opt-in' : ruleValue === 'checkbox:no-change';
-      // TODO: How to recognise inverted checkbox???
+      var isInverted = checkbox.value === 'no-change';
+      var uncheckedValue = isInverted ? 'opt-in' : 'no-change';
+      var value = checkbox.checked ? checkbox.value : uncheckedValue;
+      return ruleValue === 'checkbox:' + value;
```
katzmo: Brilliant idea!

campaignion_newsletters/src/Component.php
```diff
@@ -77,21 +77,6 @@ public function submit($email, WebformSubmission $s) {
...
-   *
-   * @return bool
-   *   TRUE if the value submitted constitutes an opt-in otherwise FALSE.
-   */
-  public function isOptIn(Submission $s) {
-    $value = $s->valuesByCid($this->component['cid']);
-    $value = ValuePrefix::remove($value);
-    return $value == 'opt-in';
-  }
-
```
katzmo, line 94: We should make a decision if we find this useful or not and consolidate with https://github.com/moreonion/campaignion/pull/111 at some point in the future.

campaignion_opt_in/src/OptInRecordFactory.php
```diff
@@ -40,7 +40,8 @@ public function recordOptIn($component, $values) {
     if (in_array($value, [Values::OPT_IN, Values::OPT_OUT])) {
       db_insert('campaignion_opt_in')->fields([
         'activity_id' => $this->activity->activity_id,
-        'operation' => reset($values),
+        'operation' => $values == Values::OPT_IN ? 1 : 0,
```
katzmo: Should be `$value == Values::OPT_IN` (without `s`)!

campaignion_opt_in/campaignion_opt_in.install
```diff
@@ -44,6 +50,7 @@ function campaignion_opt_in_schema() {
     'indexes' => [
       'activity' => ['activity_id'],
       'channel' => ['channel'],
+      'operation' => ['operation'],
```
katzmo: Good idea!

campaignion_form_builder/campaignion_form_builder.type_opt_in.inc
```diff
@@ -0,0 +1,42 @@
...
+ * Invoked by campaignion_form_builder_form_builder_element_types().
+ */
+function _campaignion_form_builder_form_builder_types_opt_in() {
+  $fields = [];
+
+  $opt_in = function($component, $weight) {
+    $element = [
+      'title' => $component['name'],
+      'weight' => $weight,
+      'unique' => TRUE,
```
katzmo, line 15: Not sure if we want it to be unique, right now, multiple newsletter fields are possible (using campaignion_newsletters)?

katzmo (changes requested):
$values → $value needs fixing, rest looks great!
(Still not completely happy with default values, but I think it’s ok for now ;))

campaignion_opt_in/src/OptInRecordFactory.php
katzmo: Pushed a fix.

katzmo (approved):

campaignion_newsletters/src/Component.php
torotil: Yes, I agree.

campaignion_form_builder/campaignion_form_builder.type_opt_in.inc
torotil: Our current interpretation of phone and post opt-ins is that they are contact-wide. So it wouldn’t make much sense to have more than one field per form, right?

katzmo: OK.

PR #102, opened 2018-04-25
Manage supporters: Only apply date filter with a complete set of values.

torotil:


torotil:
This has been pulled a long as patch for years now. At this point it’s better to just merge it than to wait for tests.

PR #103, opened 2018-04-27
3-state-optin part 2

torotil:


campaignion_newsletters/tests/ConditionalTest.php
```diff
@@ -20,44 +20,63 @@ public function setUp() {
...
+    $this->assertTrue($ne(['opt-out'], 'radios:not-selected', $radios));
+    $this->assertTrue($ne(['opt-out'], 'checkbox:opt-in', $radios));
+    $this->assertTrue($ne(['opt-out'], 'checkbox:no-change', $radios));
+    $this->assertTrue($ne(['no-change'], 'radios:opt-in', $radios));
+    $this->assertTrue($ne(['no-change'], 'radios:opt-out', $radios));
+    $this->assertFalse($ne(['no-change'], 'radios:no-change', $radios));
+    $this->assertTrue($ne(['no-change'], 'radios:not-selected', $radios));
+    $this->assertTrue($ne(['no-change'], 'checkbox:opt-in', $radios));
+    $this->assertTrue($ne(['no-change'], 'checkbox:no-change', $radios));
+    $this->assertTrue($ne(['not-selected'], 'radios:opt-in', $radios));
```
torotil: This is the test-case for testing the values directly provided by the form-API. So I guess it should still be a `[]` instead of `['not-selected']`.

torotil (commented):
Almost there I think.

campaignion_newsletters/tests/ConditionalTest.php
torotil: (Fixed that myself.)

katzmo: Your right! But this applies to the second block, testing what doesn't match, too, doesn't it?

torotil: Yes, you’re right too :-)

PR #104, opened 2018-04-27
Make links default to `node` a path that is always accessible.

torotil:
This fixes an issue where the campaignion_bar links vanish when the
front page is not accessible by the currently logged-in user.


alex-mo:
Looks good!

alex-mo (approved):

PR #106, opened 2018-05-04
Use Drupal flood protection for webform submissions.

alex-mo:
- include a patch for webform 7.x-4.16. to allows for custom limit check
  functions
- `campaignion_data_protection_webform_submission_user_limit_check` as
  limit check via Drupal core `flood` functionality
- save flood events on `webform_submission_create`


torotil:
The patch for webform belongs in the webform issue queue not here.

PR #108, opened 2018-05-07
Store remote address with newsletter subscription.

alex-mo:


torotil:
We’ll need the same for `campaignion_opt_in` once that’s in.

PR #109, opened 2018-05-07
[E2T] Update message part labels

FrClem:
Requested by: https://trello.com/c/nDOnvxLi

Missing: Add explanation message below the "Add... " buttons on /wizard/message step for email_to_target actions.


torotil:
Those are only the database column labels. This won’t have any effect on the user-visible labels. See #121

PR #118, opened 2018-06-07
Generalise "no is opt-out" option.

torotil:
- Make checkbox-inverted its own display setting.
- Allow "no is opt-out" for all displays.

### DIFF
--- a/campaignion_opt_in/src/Values.php
+++ b/campaignion_opt_in/src/Values.php
@@ -13,7 +13,7 @@ class Values {
   const NOT_SELECTED = 'not-selected';
 
   public static $noValueMap = [
-    'checkbox' => [0 => self::NO_CHANGE],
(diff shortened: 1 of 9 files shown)


campaignion_opt_in/src/Values.php
```diff
@@ -58,11 +77,81 @@ public static function addPrefix($value, $component) {
...
+
+  /**
+   * Get label for values stored in submitted data.
+   *
+   * @param mixed $values
+   *   Values as stored in the submitted data for this component.
+   */
+  public static function labelByValue($values) {
+    if (!$values) {
+      return t('Unknown value');
```
katzmo, line 114: Wouldn’t "Empty value" or "No value" be more accurat?

katzmo (approved):

campaignion_opt_in/src/Values.php
torotil: I haven’t changed that part of the logic. But why not.

PR #126, opened 2018-07-16
yarn upgrade. Remove package-lock.json.

torotil:


torotil:
~~Breaks redirect_app at the moment because of dependency issues.~~ resolved.

PR #128, opened 2018-08-10
Fix: Using isset() to field instances doesn’t work for new entities.

torotil:
campaignion_source_tag used `isset($entity->field_name)` to check whether an entity has a particular field. This does not work for new entities though. It seems the field attributes are not set on entity creation but only when the entity is first loaded from the database.


katzmo (approved):

alex-mo (approved):

torotil:
This is still waiting for a test for the redirect_app.

PR #130, opened 2018-09-04
Fixes in embed.js

katzmo:
String.prototype.match() returns `null` if there were no matches – `null` doesn’t have a length property.


alex-mo (approved):
Does the same, but more resilient

PR #132, opened 2018-09-25
Flexible selectors

torotil:
Remove postcode as hard-coded selector. Standardize handling of custom-datasets / postcode datasets.

### DIFF
--- a/campaignion_email_to_target/src/Action.php
+++ b/campaignion_email_to_target/src/Action.php
@@ -2,9 +2,10 @@
 
 namespace Drupal\campaignion_email_to_target;
 
-use \Drupal\campaignion_action\ActionBase;
-use \Drupal\campaignion_action\TypeInterface;
-use \Drupal\campaignion_email_to_target\Api\Client;
+use Drupal\campaignion_action\ActionBase;
+use Drupal\campaignion_action\TypeInterface;
+use Drupal\campaignion_email_to_target\Api\Client;
+use Drupal\little_helpers\Webform\Submission;
 
 /**
  * Defines special behavior for email to target actions.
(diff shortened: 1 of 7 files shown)


campaignion_email_to_target/src/Action.php
```diff
@@ -100,45 +130,51 @@ public function testLink($title, $query = [], $options = []) {
...
     $no_target_message = NULL;
-    foreach ($constituencies as $constituency) {
+
+    $selector = $this->buildSelector($submission_o);
+    $contacts = $this->api->getTargets($this->options['dataset_name'], $selector);
+
+    foreach ($contacts as $target) {
+      $target += ['constituency' => []];
+      $constituency = $target['constituency'];
+      unset($target['constituency']);
```
m-strasser, line 144: I don't really understand what's happening here, tbh ;)

torotil: Previously the contacts were grouped by constituency (in the API response `/v2/…/postcode/…` endpoint):

```python
[
  {'name': 'constituency1', contacts: [ {}, {}, {} ]},
  {'name': 'constituency2', contacts: [ {}, {}, {} ]},
]
```

With the new endpoint (`/v2/…/select?…`) it’s

```python
[
  {'name': 'contact1', constituency: {}},
  {'name': 'contact without constituency'},
]
```
`$contacts` in the code above is the API-response. So `$target` becomes the contact without the constituency part. And `$constituency == $target['constituency']` or an empty array if that doesn’t exist.

torotil:
```php
// This …
$target += ['constituency' => []];
// is equivalent to
if (!array_key_exists('constituency', $target)) {
  $target['constituency'] = [];
}

m-strasser: Oh okay, that explains it :)

m-strasser (approved):

PR #134, opened 2018-10-01
Allow headless nightwatch tests

alex-mo:
`redirects_app`: run tests with `HEADLESS=1 yarn e2e` for headless browser


torotil:
Tests fail for me in this branch if I invoke them as described.

alex-mo:
how do they fail? (because for me the tests are all green. I only had issues getting selenium to work, but the `HEADLESS` env should be fixing this...)

PR #136, opened 2018-10-09
[opt in] show label field for inverted checkbox component

katzmo:


torotil:
The coverage report is wrong in this case. The line is executed.

PR #138, opened 2018-10-10
campaignion_newsletters: Make conditionals not fail with ajax.

torotil:
With webform_ajax conditional rule groups accumulate in in the `Drupal.settings.webform.conditionals`. That means JS for a conditional for on one page is executed on all later pages as well. This means conditional operators need to deal with non-existent elements.

Upstream issue: https://www.drupal.org/project/webform/issues/3005684


torotil:
We have a working upstream patch!

PR #139, opened 2018-10-11
Conditionals: Support for multi-page rule groups.

torotil:
`existingValue` is used when a conditional rule group spans multiple pages. Values for components from earlier pages are then passed as `existingValue` because the `element` does not exist on this page.

*In an ideal world webform would interpret these rules on the server-side. If it were the case there would be no need for the `existingValue` parameter.*


torotil:
This is solved with [d.o#3005884](https://www.drupal.org/project/webform/issues/3005884).

PR #143, opened 2018-10-24
Backlog: save as draft (unpublish)

FrClem:
Rename "Save as draft" from the Confirm step of an action to "Save as draft (unpublish)"

German translation might need to be improved!

Request: https://trello.com/c/jhTQRAIy

### DIFF
--- a/campaignion_wizard/src/WizardStep.php
+++ b/campaignion_wizard/src/WizardStep.php
@@ -30,7 +30,7 @@ abstract class WizardStep extends _WizardStep {
     $form['buttons']['next']['#value'] = t('Next');
 
     if (isset($form['buttons']['return'])) {
-      $label = (isset($this->wizard->node->status) && $this->wizard->node->status) ? t('Save & return') : t('Save as draft');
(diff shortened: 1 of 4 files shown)


campaignion_wizard/src/WizardStep.php
```diff
@@ -30,7 +30,7 @@ public function stepForm($form, &$form_state) {
     $form['buttons']['next']['#value'] = t('Next');
 
     if (isset($form['buttons']['return'])) {
-      $label = (isset($this->wizard->node->status) && $this->wizard->node->status) ? t('Save & return') : t('Save as draft');
+      $label = (isset($this->wizard->node->status) && $this->wizard->node->status) ? t('Save & return') : t('Save as draft (unpublish)');
```
torotil, line 33: Let’s replace the `isset(…) && …` with the equivalent `!empty(…)` here.

torotil (changes requested):
Why don’t we change the label to always depend on the current `$node->status`? Then we can simply have `Save & return` and `unpublish` as actions. I don’t think the longer label works. Have you checked with Florian/Claire wrt to the labels?

torotil:
Closing this PR due to inactivity.

PR #149, opened 2018-11-09
Add auto-tagging support for utm_campaign.

torotil:


campaignion_source_tags/campaignion_source_tags.module
```diff
@@ -63,3 +71,13 @@ function campaignion_action_redhen_settings() {
   ];
   return $form;
 }
+
+/**
+ * Helper function to get the property UUID mapping.
+ */
+function _campaignion_source_tags_map() {
```
katzmo, line 78: Isn’t it rather a _campaignion_tracking_tags_map? ;)

katzmo (approved):

campaignion_source_tags/campaignion_source_tags.module
torotil: Yep. I’d have to rename the module for that though.

katzmo: Ah, right, that's the module name. Let's keep it ;)

PR #150, opened 2018-11-10
e2t: New selection mode to select a single target

torotil:


campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -53,11 +53,19 @@ function campaignion_email_to_target_menu() {
...
+  $plugins['one_or_more'] = [
+    'class' => OneOrMore::class,
+    'title' => t('At least one — Let users deselect specific targets. (checkboxes)'),
+  ];
+  $plugins['all'] = [
+    'class' => All::class,
+    'title' => t('All — Messages are sent to all targets.'),
+  ];
+  $plugins['one'] = [
     'one' => One::class,
```
katzmo: I think this should be `'class'`, not `'one'`?

torotil: True.

PR #165, opened 2019-02-28
Manage supporters: Filter for excluding archived contacts.

torotil:
The filter is enabled by default.

### DIFF
--- a/campaignion_manage/src/SupporterPage.php
+++ b/campaignion_manage/src/SupporterPage.php
@@ -2,8 +2,10 @@
 
 namespace Drupal\campaignion_manage;
 
+use Drupal\campaignion_manage\Query\Base as BaseQuery;
+
 class SupporterPage extends Page {
-  public function __construct($query) {
+  public function __construct(BaseQuery $query) {
     $this->baseQuery = $query;
 
     $filter_info = module_invoke_all('campaignion_manage_filter_info')['supporter'];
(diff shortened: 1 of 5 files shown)


alex-mo (changes requested):
`REDHEN_STATE_ARCHIVED` and `REDHEN_STATE_ACTIVE` could be used instead of `0` and `1`

alex-mo:
I guess it's ok to just exclude archived contacts and not being able to filter by status. now you cannot get only the archived ones but this might turn up as use case soon...?

campaignion_manage/tests/SupporterFiltersTest.php
```diff
@@ -0,0 +1,70 @@
...
+   * Create test contacts.
+   */
+  public function setUp() {
+    parent::setUp();
+    $c1 = Contact::fromEmail('test@example.com');
+    $c1->save();
+    $this->contacts[] = $c1;
+
+    $c2 = Contact::fromEmail('archived@example.com');
+    $c2->redhen_state = 0;
```
alex-mo: `REDHEN_STATE_ARCHIVED` :grimacing:

campaignion_manage/src/SupporterPage.php
```diff
@@ -2,15 +2,18 @@
...
+  public function __construct(BaseQuery $query) {
     $this->baseQuery = $query;
 
     $filter_info = module_invoke_all('campaignion_manage_filter_info')['supporter'];
     foreach ($filter_info as $name => $class) {
       $filters[$name] = new $class($this->baseQuery->query());
     }
-    $default[] = array('type' => 'name', 'removable' => FALSE);
+    $default[] = ['type' => 'name', 'removable' => FALSE];
+    $default[] = ['type' => 'state', 'values' => ['value' => 1]];
```
alex-mo: `REDHEN_STATE_ACTIVE`

alex-mo (changes requested):

alex-mo (approved):

PR #167, opened 2019-03-05
Vue apps unit tests

mayakokits:


.travis.yml
```diff
@@ -10,13 +14,16 @@ mysql:
   encoding: utf8
 
 cache:
+  yarn: true
```
torotil, line 17: According to https://blog.travis-ci.com/2016-11-21-travis-ci-now-supports-yarn It looks like we should also add the node_modules directories.

torotil (changes requested):
Could you rebase this so that I can merge it independently of the download/guidance PR?

.travis.yml
mayakokits: Yes maybe we should, even if the only package we're installing with npm is `yarn`. But since we install the yarn package globally, we need to cache `$/.npm`, not `node_modules`. https://twitter.com/maybekatz/status/905213355748720640
seems risky to cache all the local node_modules directories. better install them freshly with cached packages (cached via `yarn: true`).

torotil: Where are the packages stored that are installed by the individual calls to `yarn install`? Since `yarn install` is called on each and every run it’s rather safe to just cache the packages. The only danger is that removed dependencies might still be in the cache — which is a good trade-off for saving 30% of the build time.

mayakokits: yarn caches the packages in `$/.cache/yarn` which is covered by the travis shortcut `yarn: true`. when they don’t have to be downloaded any more, `yarn install` just copies them over to the local `node_modules` directories which saves quite some time. We can only guarantee proper builds if we throw away the local `node_modules` directories.

torotil: > We can only guarantee proper builds if we throw away the local `node_modules` directories.

If that’s the case then yarn is broken.

torotil: The problem is that downloading the dependencies is only about 50% of the time that’s spent on building the node_modules folder. (Just tried it for the message app on my computer.)

PR #168, opened 2019-03-11
Simplify code for triggering the email.

torotil:
The old code may have sent the email even though no submission was saved. This needs further testing.


torotil:
Ok, the `sendEmail()` method only prepares the emails for being sent by `webform_client_form_submit()`. This PR is actually breaking stuff.

PR #185, opened 2019-05-29
Refactor campaignion tracking

alex-mo:


torotil:
Where can I find the source of the bundled tracking JS? Has it changed?

campaignion_tracking/package/src/fragment.js
```diff
@@ -0,0 +1,90 @@
...
+
+/**
+ * Parse and filter a string for some URL fragment parts.
+ *
+ * @param {String[]} prefixes List of strings to filter prefixes.
+ * @param {String} locationHash The input string, typically a location.hash string.
+ *
+ * @returns {Object[]} A list of matched items.
+ */
+export function parseLocationHashForPrefixes (prefixes = [], locationHash = '') {
```
torotil, line 24: Is this actually used in production code?

campaignion_tracking/package/src/gtm.js
```diff
@@ -1,34 +1,143 @@
...
+      this.printDebug('No TrackerManager found. Doing nothing.')
+    }
+
+    /**
+     * Subscribe to fragment events.
+     *
+     * The tracker listens for specific parts in the URL fragement and
+     * published messages with event codes to trigger.
+     *
+     * See the listener.js file for this listener.
```
torotil: which listener? It’s not clear what “this listener” is referring to in this sentence.

campaignion_tracking/package/src/gtm.js
```diff
@@ -1,34 +1,143 @@
...
-}
+    if (typeof this.tracker === 'undefined') {
+      this.printDebug('No TrackerManager found. Doing nothing.')
+    }
+
+    /**
+     * Subscribe to fragment events.
+     *
+     * The tracker listens for specific parts in the URL fragement and
+     * published messages with event codes to trigger.
```
torotil: Why does it need to listen for both? Isn’t the TrackerManager supposed to turn fragments and on-page events into a unified source of pubsub-messages?

campaignion_tracking/package/src/main.js
```diff
@@ -11,13 +12,14 @@ import * as listener from './listener'
 var debug = parseInt(sessionStorage.getItem('campaignion_debug')) ? true : false
 
 // some out-of-the-box prefixes
-export const codePrefixes = []
+export const codePrefixes = ['gtm']
```
torotil: Why are there tracker specific fragment prefixes?

campaignion_tracking/campaignion_tracking.module
```diff
@@ -13,3 +22,193 @@ function campaignion_tracking_preprocess_html(&$vars) {
...
+    else {
+      $form_state['redirect'][1]['fragment'] = $fragment;
+    }
+  }
+}
+
+/**
+ * Load a payment.
+ *
+ * TODO: is there a little_helper for that?
```
torotil: No. It’s hardly the responsibility of little_helpers. You might be interested in [`$submission->payments` provided by webform_paymethod_select](https://github.com/moreonion/webform_paymethod_select/blob/530b79a9d1dfa6710f1bc681229ccdc65221b76a/webform_paymethod_select.module#L392-L436)

You might need [this workaround from campaignion_logcrm](https://github.com/moreonion/campaignion_logcrm/blob/0869ba2089ecc6fac9004ab5ff9078056ba36afe/integrations/webform_paymethod_select.inc#L15-L20) though.

campaignion_tracking/campaignion_tracking.module
```diff
@@ -13,3 +22,193 @@ function campaignion_tracking_preprocess_html(&$vars) {
...
+}
+
+/**
+ * Implements hook_node_view().
+ *
+ * Information from the previous node, where a webform was submitted.
+ *
+ * TODO: Maybe also do this in node_view
+ */
+function campaignion_tracking_page_build(&$page) {
```
torotil: Shouldn’t this be made obsolete by the fragment appended to the redirect URL?

campaignion_tracking/js/campaignion_tracking.js
```diff
@@ -0,0 +1,123 @@
...
+          donation['amount'] &&
+          donation['interval']) {
+        var product = {
+          name: node['title'],
+          id: node['nid'],
+          price: String(donation['amount']),
+          variant: String(donation['interval']),
+          quantity: 1
+        };
+        // TODO change name, place
```
torotil: I don’t understand what this `TODO` wants to tell me.

campaignion_tracking/js/campaignion_tracking.js
```diff
@@ -0,0 +1,123 @@
...
+   * available. They are only fired once for the Drupal context to prevent the
+   * same messages being dispatch multiple times (behaviours can be called
+   * multiple times).
+   *
+   * Events:
+   * - 'setDonationProduct'
+   * - 'checkoutBegin'
+   * - 'checkoutEnd'
+   */
+  $('.webform-client-form', context).once('campaignion-tracking', function () {
```
torotil: The `once()` hinders any event from being dispatched that is not in the `settings` the first time the behaviour is called. I don’t think this is the intended behaviour.

torotil (changes requested):
That’s a big one. Let’s huddle on Monday.

campaignion_tracking/package/src/fragment.js
alex-mo: not yet.

campaignion_tracking/package/src/gtm.js
alex-mo: well, there is only one listener in the referenced file, but yes the FragmentListener is the one referenced to. Could be more explicit...

campaignion_tracking/package/src/gtm.js
alex-mo: No. The TrackerManager currently should only turn fragments into code messages. It does not care what they mean, just adds an prefix. The interested trackers themselves are responsible to listen to them. So they can decide not to be interested in them.

Could be changed of course...

campaignion_tracking/package/src/main.js
alex-mo: The reason was to be easily integratable with the facebook pixel code which is working that way.

campaignion_tracking/campaignion_tracking.module
alex-mo: will take a look!

campaignion_tracking/js/campaignion_tracking.js
alex-mo: ah, yes, this was referring to a prior code version. was an annotation to think again about the domain language and where to put the creation of the `product` object)

campaignion_tracking/js/campaignion_tracking.js
alex-mo: this is called once per page load (when not using webform AJAX) or once per AJAX form submit.
AJAX form submits completely replace the `<form>` markup, thus this handler get's called exactly once per (successful) submit on multistep forms.

The reason for this once is that otherwise the behaviour get's called *multiple* times *per* 1 submit.

Maybe there is a better way. But it does what it is supposed to to (as far as I have tested)...

campaignion_tracking/package/src/fragment.js
alex-mo: is now.

campaignion_tracking/package/src/gtm.js
alex-mo: move the code listener to the common module for the Drupal integration and changed the messages chnnel to just `code` (not eg. `code:gtm`) any more...

campaignion_tracking/campaignion_tracking.module
alex-mo: gone now.

the fragment triggers a message on the trackermanager and the gtm tracker uses it's sessionStorage state to get the information (along with additional information from the fragment like the sid and paymethod)

campaignion_tracking/js/campaignion_tracking.js
alex-mo: reworked. `once()` is gone now. introduced `settings['campaignion_tracking']['sent']` to be sure we are not sending the same events twice...

campaignion_tracking/campaignion_tracking.module
```diff
@@ -0,0 +1,175 @@
...
+
+use Drupal\campaignion_action\Loader;
+use Drupal\little_helpers\Webform\Submission;
+use Drupal\little_helpers\Webform\Webform;
+
+/**
+ * Implements hook_preprocess_html().
+ */
+function campaignion_tracking_preprocess_html(&$vars) {
+  $module_path = drupal_get_path('module', 'campaignion_tracking');
```
torotil, line 16: In general I prefer `#attached` over `drupal_add_*`. This would require moving this into `hook_page_build()` or something similar.

campaignion_tracking/campaignion_tracking.module
```diff
@@ -0,0 +1,175 @@
...
+
+/**
+ * Implements hook_node_view().
+ */
+function campaignion_tracking_node_view($node, $view_mode, $langcode) {
+  $type = Loader::instance()->type($node->type);
+  $is_donation = $type && $type->isDonation();
+
+  // General information about the node.
+  drupal_add_js([
```
torotil: This should use: `$node->content['#attached']` (or something similar).

campaignion_tracking/campaignion_tracking.module
```diff
@@ -0,0 +1,175 @@
...
+ * A submit handler is added to deal with redirects.
+ */
+function campaignion_tracking_form_webform_client_form_alter(&$form, &$form_state, $form_id) {
+  $node = $form['#node'];
+  $type = Loader::instance()->type($node->type);
+  $is_donation = $type && $type->isDonation();
+  $webform = Webform::fromNode($form['#node']);
+
+  // Information about the current webform.
+  drupal_add_js([
```
torotil: Please use `$form['#attached']` here instead of `drupal_add_js()`.

campaignion_tracking/campaignion_tracking.module
```diff
@@ -0,0 +1,175 @@
...
+    $form['#submit'][] = 'campaignion_tracking_add_donation_fragment';
+  }
+}
+
+/**
+ * Submit-callback for webform_client_form().
+ *
+ * Add webform tracking data to the URL-fragment of the redirect.
+ */
+function campaignion_tracking_add_webform_fragment($form, &$form_state) {
```
torotil: This should be done using `hook_webform_redirect_alter()` defined in `little_helpers` — otherwise it won’t work with custom redirects.

campaignion_tracking/campaignion_tracking.module
```diff
@@ -0,0 +1,175 @@
...
+    $form_state['redirect'][1]['fragment'] = $fragment;
+  }
+}
+
+/**
+ * Submit-callback for webform_client_form().
+ *
+ * Add donation success tracking data to the URL-fragment of the redirect.
+ */
+function campaignion_tracking_add_donation_fragment($form, &$form_state) {
```
torotil: Also needs to be done in `hook_webform_redirect_alter()` to work with custom redirects.

campaignion_tracking/js/campaignion_tracking.js
```diff
@@ -0,0 +1,82 @@
...
+   * - 'checkoutEnd'
+   */
+  if (settings['campaignion_tracking'] && settings['campaignion_tracking']['context']) {
+    settings['campaignion_tracking']['sent'] = settings['campaignion_tracking']['sent'] || [];
+
+    var node = settings.campaignion_tracking.context['node'] || {};
+    var donation = settings.campaignion_tracking.context['donation'] || {};
+    var webform = settings.campaignion_tracking.context['webform'] || {};
+
+    if (node['is_donation']) {
```
torotil, line 45: Would it be better to dispatch the events during the submit-events on the form? — Maybe something to be discussed broken out into another ticket.

campaignion_tracking/package/src/drupal-common.js
```diff
@@ -0,0 +1,84 @@
...
+// t: tracking, w: webform, d: donation
+export const codePrefixes = ['t', 'w', 'd']
+
+// common tracker manager, listener, gtm
+export const tracker = new tm.TrackerManager(debug)
+export const fragmentListener = new listener.FragmentListener(tracker, codePrefixes, debug)
+fragmentListener.setup()
+
+/**
+ * Subscribe to messages of the `code:gtm` tracking channel.
```
torotil: The comment seems to be outdated (still referring to `gtm`).

campaignion_tracking/package/src/gtm.js
```diff
@@ -0,0 +1,335 @@
...
+   */
+  callChangeHook (eventName, gtmData, context) {
+    // maybe also inject tracker
+    if (typeof window['campaignion_tracking_change_msg'] === 'function') {
+      gtmData = window.campaignion_tracking_change_msg('gtm', eventName, gtmData, context)
+    }
+    return gtmData
+  }
+
+  updateContext (context) {
```
torotil, line 154: How is this context different from the tracker independent context? What additional information does it contain?

torotil (changes requested):
Much better already!

* Happy to discuss my questions ;-)
* Still needs a bit of coverage for the PHP changes I guess.

campaignion_tracking/package/src/gtm.js
alex-mo: Not a all. It's here mainly because there was not need (yet) to have it in the tracker independent place... :)

So it's missing refactoring...

campaignion_tracking/campaignion_tracking.module
alex-mo: Is there a reason to prefer `#attached`?

torotil: Yes. It makes your function a pure function instead of having side-effects, which means you might be able to actually test it.

campaignion_tracking/js/campaignion_tracking.js
alex-mo: yes, agree, would be better. also yes, maybe best in a new card...

campaignion_tracking/package/src/drupal-common.js
alex-mo: fixed

campaignion_tracking/campaignion_tracking.module
alex-mo: If that is the goal, then you cannot use `$element['#attached']['js'][]` but need define a key, I guess. Otherwise it's not pure...

campaignion_tracking/campaignion_tracking.module
alex-mo: ok

campaignion_tracking/campaignion_tracking.module
torotil: You should be able to translate this directly into:

```php
$element['#attached'][$path] = $options;
```

See [drupal_process_attached()](https://devdocs.io/drupal~7/includes-common.inc/function/drupal_process_attached/7.x)

campaignion_tracking/campaignion_tracking.module
alex-mo: changed it

campaignion_tracking/campaignion_tracking.module
alex-mo: changed

campaignion_tracking/campaignion_tracking.module
alex-mo: changed

campaignion_tracking/campaignion_tracking.module
alex-mo: changed

torotil (approved):

PR #188, opened 2019-06-07
Vue apps dependency updates

mayakokits:
Uprade vue and element-ui.


mayakokits:
To resolve the conflict, just use the `datasets_app.vue.min.js` from the `vue-apps-dependency-updates` branch when merging. I’ve recompiled it so it features the tooltip added by Michael (25d11fc).

PR #189, opened 2019-06-19
Allow multiple email opt-ins. Add interest group tab.

katzmo:


torotil (approved):

torotil (changes requested):
Ok, that breaks mailchimp interest groups at the moment.

torotil (approved):

PR #194, opened 2019-08-27
fix function declaration error in php 7.3

mojulia:


src/CRM/Export/DateField.php
```diff
@@ -8,7 +8,7 @@ public function __construct($key, $format) {
     parent::__construct($key);
     $this->format = $format;
   }
-  public function value() {
+  public function value($delta = 0) {
     if ($timestamp = parent::value()) {
```
torotil: `$delta` should be passed to the parent here.

src/CRM/Export/KeyedField.php
```diff
@@ -9,8 +9,8 @@ public function __construct($key, $subkey) {
     $this->subkey = $subkey;
   }
 
-  public function value() {
-    if (($values = parent::value(NULL))) {
+  public function value($delta = NULL) {
+    if (($values = parent::value($delta))) {
```
torotil: The data structure returned by `WrapperField::value()` differs depending on whether you pass `NULL` or something else (ie. the default `0`). So passing anything else than `NULL` to this function probably wouldn’t work.

torotil (changes requested):

torotil (approved):

katzmo (approved):

PR #207, opened 2019-11-11
Don’t require unique email addresses for forms with captcha.

torotil:


campaignion_wizard/src/WebformStepUnique.php
```diff
@@ -30,4 +30,26 @@ public function validateStep($form, &$form_state) {
...
+   *   The node which’s captcha settings should be checked.
+   *
+   * @return bool
+   *   TRUE if a captcha is set up for this node’s webform otherwise FALSE.
+   */
+  protected function hasCaptcha($node) {
+    if (!module_exists('captcha')) {
+      return FALSE;
+    }
+    $form_id = "webform_client_form_{$node->nid}";
```
katzmo, line 47: Is that save? What if there are several forms on the same nid?

campaignion_wizard/src/WebformStepUnique.php
```diff
@@ -20,7 +20,7 @@ public function validateStep($form, &$form_state) {
         $c = reset($cs);
       }
     }
-    if ($c) {
+    if ($c && !$this->hasCaptcha($form['#node'])) {
```
katzmo, line 24: I don’t quite get the logic behind "if there is a capture, the email address does not have to be unique", aren’t those completely different things? For spamming a form, one could easily generate unique email addresses. On the other hand a capture won’t stop real people from submitting the form twice with the same data.

torotil: True. Anyway we require only at least one of them to be set up even if that’s not a comprehensive solution.

campaignion_wizard/src/WebformStepUnique.php
torotil: How can there be more than one webform for a webform node?

katzmo: Well if there can’t, it’s fine ;)

katzmo (approved):

PR #215, opened 2020-02-12
[opt_in] Load opt-ins with submissions.

katzmo:


campaignion_opt_in/campaignion_opt_in.module
```diff
@@ -132,6 +132,27 @@ function campaignion_opt_in_campaignion_action_taken($node, Submission $submissi
...
+ * Add opt-ins to their submissions. The opt-ins will be accessible
+ * under $submission->optins[$cid].
+ */
+function campaignion_opt_in_webform_submission_load(&$submissions) {
+  foreach ($submissions as $sid => $submission) {
+    $submission->optins = [];
+  }
+  $result = db_query('SELECT sid, channel, operation, value, statement, ip_address FROM campaignion_opt_in INNER JOIN campaignion_activity_webform USING(activity_id) WHERE sid in (:sids)', [':sids' => array_keys($submissions)]);
+  foreach ($result as $optin_data) {
+    $submissions[$optin_data->sid]->optins[$optin_data->channel] = [
```
torotil: What if there are multiple opt-in components per channel in one submission?

katzmo: ~~They would still trigger only one opt-in activity per channel.~~ (Not sure what happens if an opt_in component and a phone_number component in the same form both collect a 'phone' opt-in, I guess this would trigger 2 different activites for the same channel but probably shouldn't?)

katzmo: Addressed in  https://github.com/moreonion/campaignion/pull/215/commits/1494ec3e27547393c7aa5399b4d41d63ad36d66d

torotil:
Closing this in favor of #216 .

PR #218, opened 2020-02-20
IE11 compatibility

alex-mo:


torotil:
The current plan is to push the responsibility for loading the necessary polyfills to downstream. We’ll still transpile for IE11 / ES5 for the time being but it’s not a sensible approach to add polyfills in each and every JS bundle. So this becomes more of a documentation issue.

PR #222, opened 2020-03-06
Tracking "submission" events.

alex-mo:


campaignion_tracking/campaignion_tracking.module
```diff
@@ -112,12 +118,14 @@ function campaignion_tracking_webform_redirect_alter($redirect, $submission) {
...
-      // Prefix `dm` for donation method.
-      $fragment .= ";d:m=" . $payment->method->name;
+      // Add context for donation tracking.
+      $fragments[] = "d:m=" . $payment->method->name;
     }
   }
 
-  $redirect->fragment = empty($redirect->fragment) ? $fragment : $redirect->fragment . ';' . $fragment;
+  $tracking_fragment = implode(';', $fragments);
+  $redirect->fragment = empty($redirect->fragment) ? $tracking_fragment : $redirect->fragment . ';' . $tracking_fragment;
```
torotil, line 137: How does this work wrt url encoding? Why are the values individually prefixed instead of using `&`-separated key value pairs?

torotil (changes requested):
Apart from the question in the other comment:

- Yes, this needs indeed rebasing.
- ~~The recompile JS commit shouldn’t be included in the PR.~~
- Tests for the PHP-part are failing, they also might need additional coverage.

campaignion_tracking/campaignion_tracking.module
alex-mo: Is there a convention to use `&` in our fragment encoding?

alex-mo: Can you give an example how you would suggest to encode the information in the fragment?

torotil: The idea was to have something like this:

```
#real-fragment-without-namespace;namespace1:key1=value1&key2=value2;namespace2:key1=value1&…
```

alex-mo:
- Rebased
- Removed compiled assets

campaignion_tracking/campaignion_tracking.module
alex-mo: I see.

Although there is not much difference besides brevity to
```
#real-fragment-without-namespace;namespace1:key1=value1;namespace1:key2=value2;namespace2:key1=value1;…
```

- Without `&` parsing/encoding is simpler.
- Namespaces in this module are quite short (`t`, `d`, `w`), so not much added length
- Concepts stay the same

alex-mo: We probably can just add splitting by `&` to the parser. If namespaces appear multiple times, they just get merged. This is currently how the tracking code work. Easier to just add additional information then...

torotil: This scheme was invented with webform_prefill. I’d like to keep it consistent across projects. If the tracking code works differently then this is a bug.

torotil: I don’t think parsing is simpler: This way parsing is.

1. split by `;`
2. Find parts that start with the right `prefix:`.
3. Use a parser for URL-parameters (usually in the standard library) to parse the string and get all key value pairs and decode them.

campaignion_tracking/package/src/gtm.js
```diff
@@ -154,11 +171,59 @@ export class GTMTracker {
...
+    let data = {
+      event: 'submission',
+      webform: {
+        nid: eventData.nid || null,
+        sid: eventData.sid || null,
+        title: eventData.title || null,
+      }
+    }
+    // Allow others to modify the data being sent to GTM.
+    data = this.callChangeHook(eventName, data, this._context)
```
torotil, line 223: Is there a reason to implement this in hook-style as opposed to JS events?

campaignion_tracking/package/src/fragment.js
```diff
@@ -57,35 +63,40 @@ export function consumeLocationHashForPrefixes (prefixes = [], locationHash = ''
...
+ *
+ * @param {String} part Parts from the URL fragment, delimited with `&`.
+ * @param {String} prefix A prefix to tag the parts with
+ *
+ * @returns {Array} A list of match objects
+ */
+export function parsePartsWithPrefix (partsString, prefix = '') {
+  const split = partsString.split('&')
+  const parts = split.map(part => {
+    const _p = parsePart(part)
```
torotil, line 97: Does this handle URL-encoding? (I’m not sure where’s the best place to add the comment)

campaignion_tracking/campaignion_tracking.module
```diff
@@ -119,13 +120,19 @@ function campaignion_tracking_webform_redirect_alter($redirect, $submission) {
...
-      $fragments[] = "d:m=" . $payment->method->name;
+      $donationFragments[] = "m=" . $payment->method->name;
     }
   }
 
+  $fragments[] = 't:'. implode('&', $trackingFragments);
+  $fragments[] = 'w:'. implode('&', $webformFragments);
+  if ($donationFragments) {
+    $fragments[] = 'd:'. implode('&', $donationFragments);
+  }
```
torotil, line 134: Why do we use 3 prefixes for a single module?

torotil (commented):
Still a few questions … Nothing that I consider blockers though.

campaignion_tracking/campaignion_tracking.module
alex-mo: this was the original design which I left untouched for the moment... idea was to have different prefixes for different "sources"...

happy to use just one `t` namespace as well if this is more appreciated :)

campaignion_tracking/package/src/gtm.js
alex-mo: how would you receive back the modified data then? :thinking: 

but I agree, there might be opportunity for refactoring here...

campaignion_tracking/package/src/fragment.js
alex-mo: Should be not an issue wrt `&`, they are not escaped:

```
> window.location.hash = '& '
"& "
> window.location.hash
"#&%20"
```

(for Fx and Chroumium)

campaignion_tracking/campaignion_tracking.module
alex-mo: Should we refactor this now? (I guess now is better than "later"...)

torotil: How about `w:title=One: & two; & three,`?

campaignion_tracking/package/src/fragment.js
torotil: The parser still needs to handle those values properly and pass them *decoded* to the events.

How about `w:title=One: & two; & three,`?

campaignion_tracking/campaignion_tracking.module
torotil: I’m not sure whether now is the best time. The original intention was to give each module (or use-case) one prefix so that they don’t interfere with each other.

campaignion_tracking/package/src/fragment.js
alex-mo: OK, thanks, got it...

alex-mo: Should be fixed now.
- Assuming all values need to be encoded
- encoding only relevant string here: `title` (the other values are hardcoded chars or integers)

torotil (approved):

PR #223, opened 2020-03-10
E2t message token

alex-mo:


campaignion_email_to_target/campaignion-email-to-target-all-messages-mail.tpl.php
```diff
@@ -0,0 +1,22 @@
...
+ * Available variables:
+ * - $messages: An array of mails sent to targets.
+ * - $submission: The submission of the email to target action.
+ */
+?>
+
+<?php foreach ($messages as $message) : ?>
+<p>Email was sent to: <?php echo htmlspecialchars($message['to']); ?></p>
+
+<div>
```
torotil: I would also include the subject line in the template.

campaignion_email_to_target/campaignion_email_to_target.info
```diff
@@ -5,6 +5,7 @@ package = Campaignion
 
 dependencies[] = campaignion_action
 dependencies[] = campaignion_vue
+dependencies[] = campaignion_webform_token
```
torotil: Why is this dependency added?

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -153,35 +155,70 @@ function campaignion_email_to_target_token_info() {
       drupal_set_message(t('We are sorry! The service is temporary unavailable. The administrators have been informed. Please try again in a few minutes …'), 'error');
     }
   }
+
+  $info['tokens']['email-to-target']['messages'] = [
```
torotil: If we put the token in the `submission` group it would be automatically available on thank you pages for example. I’d reserve the `email-to-target` tokens for stuff that can be replaced inside of and per message.

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -153,35 +155,70 @@ function campaignion_email_to_target_token_info() {
...
+  $replacements = [];
+  if ($type != 'email-to-target') {
+    return $replacements;
   }
 
-  $my_data['contact'] = $data['email-to-target'];
-  foreach ($my_data['contact'] as $key => $sub_array) {
-    if (is_array($sub_array)) {
-      $my_data[$key] = $sub_array;
+  // Token for complete messages sent to target.
```
torotil: I’d suggest splitting the token callback function into two: one for `submisson` tokens and one for the `email-to-target` tokens. We could put them into 1 or 2 new classes.

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -153,35 +155,70 @@ function campaignion_email_to_target_token_info() {
...
+  if (!empty($tokens['messages']) && !empty($data['webform-submission']) && !empty($data['node'])) {
+    $submission = new Submission($data['node'], $data['webform-submission']);
+    $action = Loader::instance()->actionFromNode($submission->node);
+    $channel = $action->channel();
+    $components = $submission->webform->componentsByType('e2t_selector');
+    $messages = [];
+    foreach ($components as $cid => $component) {
+      $component_o = Component::fromComponent($component);
```
torotil: `$component_o` is not used as far as I can see.

torotil (changes requested):
General approach looks good.

I’m not sure about the `prepareEmail()` idea. I think I’d prefer a theme function that directly gets `$message` and `$submission` — although this might mean that we invoke emai theme function from two places. Then we could also add the channel as theme-hook-suggestion so that different channels can be easily rendered differently. If the channel-concept is introduced more thoroughly it’s likely that the `$message` object will look differently depending on the channel.

campaignion_email_to_target/campaignion_email_to_target.info
alex-mo: ah, yes, not needed any more.
I used to rely on `_campaignion_webform_tokens_get_submission` before realising that it's already in the token data... :innocent:

campaignion_email_to_target/src/Tokens.php
```diff
@@ -0,0 +1,66 @@
+<?php
+
+namespace Drupal\campaignion_email_to_target;
+
+use Drupal\campaignion_action\Loader;
```
alex-mo: Not used any more.

alex-mo (commented):
Your ideas of theme hook suggestions based on the channel are left out on purpose, I assume. We probably add them when needed, right?

torotil:
> Your ideas of theme hook suggestions based on the channel are left out on purpose, I assume. We probably add them when needed, right?

Yes It’s hard to test them … and if it comes to that `Messages` will need to change as well it seems. So I’ve postponed that.

torotil (approved):

PR #226, opened 2020-04-14
Fixing typo. Is taged with => Is tagged with

edjw:


torotil (approved):

torotil:
Thanks! This will be part of the next release which is scheduled for 6/5.

PR #229, opened 2020-05-01
e2t: Remove NullCacheBin when disabling the module.

torotil:
The problem was made visible with the latest little_helpers changes that deactivated autoloading modules from disabled/uninstalled modules.

### DIFF
--- a/campaignion_email_to_target/campaignion_email_to_target.install
+++ b/campaignion_email_to_target/campaignion_email_to_target.install
@@ -7,6 +7,7 @@
 
 use Drupal\campaignion_email_to_target\Api\Client;
 use Drupal\campaignion_email_to_target\Api\ConfigError;
+use Drupal\campaignion_email_to_target\NullCacheBin;
 
 /**
  * Implements hook_schema()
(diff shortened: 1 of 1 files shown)


campaignion_email_to_target/campaignion_email_to_target.install
```diff
@@ -138,7 +139,16 @@ function campaignion_email_to_target_field_schema($field) {
...
 function campaignion_email_to_target_enable() {
-  variable_set('cache_class_cache_token', '\\Drupal\\campaignion_email_to_target\\NullCacheBin');
+  variable_set('cache_class_cache_token', NullCacheBin::class);
+}
+
+/**
+ * Implements hook_disable().
+ */
+function campaignion_email_to_target_disable() {
+  if (variable_set('cache_class_cache_token') == NullCacheBin::class) {
```
katzmo: Shouldn't this be `variable_get()`?

torotil: True. Thanks!

katzmo (approved):

PR #237, opened 2020-06-29
Make language switcher more themable.

torotil:
- Avoid `theme_links()` for easier theming.

### DIFF
--- a/campaignion_language_switcher/campaignion-language-switcher.tpl.php
+++ b/campaignion_language_switcher/campaignion-language-switcher.tpl.php
@@ -1,0 +1,21 @@
+<?php
+
+/**
+ * @file
+ * Display a list of translation links.
+ *
+ * Available variables:
+ * - $links: An array of $link-arrays. Each $link in $links contains:
+ *   - $link['li_attributes']: Attribute array for the list-item.
+ *   - $link['renderable']: A renderable array for the link content.
+ *
+ * @see template_preprocess_campaignion_language_switcher()
+ *
+ * @ingroup themeable
+ */
+?>
+<ul class="campaignion-language-switcher-<?php echo $provider; ?>">
+<?php foreach ($links as $link): ?>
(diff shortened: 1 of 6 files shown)


campaignion_language_switcher/campaignion_language_switcher.theme.inc
```diff
@@ -0,0 +1,63 @@
...
+ *   An associative array containing:
+ *   - links: An array of links similar to theme_links().
+ *   - provider: Machine name of a translation provider.
+ */
+function template_preprocess_campaignion_language_switcher(array &$variables) {
+  global $language_url;
+
+  $links = &$variables['links'];
+  $num_links = count($links);
+  $i = 1;
```
alex-mo: These seem not to be used...

campaignion_language_switcher/campaignion-language-switcher.tpl.php
```diff
@@ -0,0 +1,21 @@
...
+ * - $links: An array of $link-arrays. Each $link in $links contains:
+ *   - $link['li_attributes']: Attribute array for the list-item.
+ *   - $link['renderable']: A renderable array for the link content.
+ *
+ * @see template_preprocess_campaignion_language_switcher()
+ *
+ * @ingroup themeable
+ */
+?>
+<ul class="campaignion-language-switcher-<?php echo $provider; ?>">
```
alex-mo, line 18: Wrapping it in a `<nav class="campaignion-language-switcher">` would be better markup (after all, it has a menu role), but would break backward compatability with current themes, I guess...

alex-mo (changes requested):

campaignion_language_switcher/campaignion-language-switcher.tpl.php
torotil: Yep, there are many themes with `ul.campaignion-language-switcher` selectors. New themes can now easily override the template.

alex-mo (approved):

katzmo (approved):

PR #238, opened 2020-07-14
[newsletters] Fix inconsistent variable names

katzmo:
Alternatively, we could also change the variable read in [Component::fromComponent](https://github.com/moreonion/campaignion/blob/7.x-2.x/campaignion_newsletters/src/Component.php#L41) but this seems safer as it keeps the variable that’s potentially in use now (and working) unchanged.

### DIFF
--- a/campaignion_newsletters/campaignion_newsletters.install
+++ b/campaignion_newsletters/campaignion_newsletters.install
@@ -196,6 +196,16 @@ function campaignion_newsletters_uninstall() {
 }
 
 /**
+ * Rename variable.
+ */
+function campaignion_newsletters_update_21() {
+  if ($old_value = variable_get_value('campaignion_newsletters_opt_out_unknown')) {
+    variable_set_value('campaignion_newsletters_unsubscribe_unknown', $old_value);
+  }
+  variable_del('campaignion_newsletters_opt_out_unknown');
+}
+
+/**
  * Migrate to using opt_in components.
  */
 function campaignion_newsletters_update_20() {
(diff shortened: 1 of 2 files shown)


campaignion_newsletters/campaignion_newsletters.variable.inc
```diff
@@ -23,7 +23,7 @@ function campaignion_newsletters_variable_info($options) {
     'default' => 5,
     'localize' => FALSE,
   ];
-  $v['campaignion_newsletters_unsubscribe_unknown'] = [
```
torotil: I think the naming is somewhat consistent:

- `campaignion_opt_in` talks about “opt-ins” and “opt-outs”.
- `campaignion_newsletters` talks about subscriptions, subscribe and unsubscribe.

Why do you think it is inconsistent?

katzmo: Because `campaignion_newsletters_unsubscribe_unknown` is not used by any code, instead `variable_get()` [here](https://github.com/moreonion/campaignion/blob/7.x-2.x/campaignion_newsletters/src/Component.php#L41) reads `campaignion_newsletters_opt_out_unknown`.

torotil: Ok, then IMHO the variable read is at fault.

katzmo: Changed the pull request. But now we’ll need a migration for already existing variables…

torotil: True. So an update hook is needed, right?

torotil (approved):
The migration will lose explicitly set `FALSE` value. Since there is no reason to actually have those (and we don’t have them in any site) this solution is good for all real world cases.

PR #240, opened 2020-07-21
WIP: Campaignion donation amount total token

mojulia:
please review, especially tests need help!


torotil:
This looks very much installation specific. This should be put in the profile instead until we figure out a way to make this more generic.

PR #244, opened 2020-08-16
Donor segmentation fields

torotil:


campaignion_donation/campaignion_donation.module
```diff
@@ -125,3 +126,47 @@ function campaignion_donation_form_builder_element_types_alter(&$types, $form_ty
...
+        continue;
+      }
+      $field_name = $recurrent ? 'donation_latest_regular' : 'donation_latest_one_off';
+      if (isset($contact->{$field_name})) {
+        $field = $contact->wrap()->{$field_name};
+        // If there is no field_collection_item yet, we need to create one.
+        if (!$field->value()) {
+          entity_create('field_collection_item', ['field_name' => $field_name])
+            ->setHostEntity('redhen_contact', $contact);
+          // $field->set() does not at the host entity field.
```
katzmo: at → add?

campaignion_donation/campaignion_donation.features.field_base.inc
```diff
@@ -0,0 +1,117 @@
...
+    'module' => 'number',
+    'settings' => array(
+      'decimal_separator' => '.',
+    ),
+    'translatable' => 0,
+    'type' => 'number_float',
+  );
+
+  return $field_bases;
+}
```
katzmo, line 117: I’m wondering how useful it is to have a donation amount without currency and a recurring amount without interval? Shouldn’t they be part of the fieldcollection (or the donation amount) too?

torotil: Only the donation amount was required in the specs so … I guess even that in itself has value.

I agree that having currency and frequency would be even better.

campaignion_donation/campaignion_donation.module
torotil: fixed. thanks!

katzmo (approved):

PR #248, opened 2020-10-05
Donation amount component

torotil:
Aim is to improve the usability of configuring donation amount components. In order to do this we add a new customized webform component.

1. Pre-defined values are entered as numbers (no need to care about the mapping of the label to the value).
2. Functionality is based on the number webform component so it has the same functionality
   - Setting values using conditionals is possible
   - Configuration of the validation is possible (min / max amount)
   - Numeric operators for conditionals.


campaignion_donation/components/donation_amount.php
```diff
@@ -0,0 +1,213 @@
...
+/**
+ * Implements _webform_defaults_component().
+ */
+function _webform_defaults_donation_amount() {
+  $defaults = webform_component_invoke('number', 'defaults');
+  unset($defaults['type']);
+  $defaults['extra'] += [
+    'options' => [],
+    'other_option' => TRUE,
+    'other_text' => t('Other'),
```
alex-mo: Why not "Other amount"? Seems a better fit as it can only be another amount, not another "something".

campaignion_donation/components/donation_amount.php
```diff
@@ -0,0 +1,213 @@
...
+  ];
+
+  return $form;
+}
+
+/**
+ * Element validate callback: Split amount array.
+ */
+function _webform_edit_donation_amount_validate_amounts($element, &$form_state) {
+  $values =
```
alex-mo: This line does not seem to be terminated... (or I do not know the syntax/code style for this code...)

alex-mo: But as `$values` is set further down below I guess this line can be removed...

campaignion_donation/components/donation_amount.php
```diff
@@ -0,0 +1,213 @@
...
+function _webform_render_donation_amount($component, $value = NULL, $filter = TRUE, $submission = NULL) {
+  $number_component = $component;
+  $component['extra']['type'] = 'textfield';
+  $component['extra']['unique'] = FALSE;
+  $element = webform_component_invoke('number', 'render', $component, $value, $filter, $submission);
+  $element['#attributes']['class'][] = 'donation-amount';
+  if ($component['extra']['options']) {
+    $element['#type'] = 'select_or_other';
+    $element['#select_type'] = 'radios';
+    $element['#options'] = drupal_map_assoc($component['extra']['options']);
```
alex-mo: Would be great if we can toggle whether the currency symbol gets prepended before the amount. This is a very common use case in rendering. (Just rendering it an hiding via styles would work too)

The currency symbol would be best wrapped in a span with an appropriate class...

campaignion_donation/components/donation_amount.php
```diff
@@ -0,0 +1,213 @@
...
+    $element['#other'] = !empty($component['extra']['other_text']) ? check_plain($component['extra']['other_text']) : t('Other...');
```
alex-mo: Same as above: why not "Other amount"?

PR #249, opened 2020-10-09
Donation tracking: Use the first non-empty line-item.

alex-mo:
It should work with 2 different, conditionally rendered amount fields

### DIFF
--- a/campaignion_tracking/campaignion_tracking.module
+++ b/campaignion_tracking/campaignion_tracking.module
@@ -75,17 +75,31 @@ function campaignion_tracking_form_webform_client_form_alter(&$form, &$form_stat
 
   // Information about a donation (in case the node is one).
   if ($is_donation) {
+    // TODO: there might be more than 1 paymethod_select component in the form
     $paymethod = $webform->componentByKey('paymethod_select');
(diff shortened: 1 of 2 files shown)


campaignion_tracking/campaignion_tracking.module
```diff
@@ -75,17 +75,31 @@ function campaignion_tracking_form_webform_client_form_alter(&$form, &$form_stat
...
     // Thus `interval` and `amount` might be "null" in JS (in case no values
     // have been submitted so far)
     $donation_settings['campaignion_tracking']['context']['donation'] = [
       'currency_code' => $currency_code,
-      'interval' => $submission->valueByKey('donation_interval'),
-      'amount' => $submission->valueByKey('donation_amount'),
+      'interval' => $interval,
+      'amount' => $selected_line_item ? $submission->valueByCid($selected_line_item->amount_component) : NULL,
+      'amount_component' => $selected_line_item ? $selected_line_item->amount_component : NULL,
+      'description' =>$selected_line_item ? $selected_line_item->description : NULL,
```
torotil: A space is missing after `=>`.

torotil (commented):
Let’s keep in mind that this still only works for one line item (with an amount). The extension is very specific to the current use-case.

torotil (approved):

PR #251, opened 2020-10-21
e2t: Make message form themable. Render non-editable subject as text.

torotil:


campaignion_email_to_target/campaignion-email-to-target-email-message-form.tpl.php
```diff
@@ -0,0 +1,31 @@
...
+if ($element['#editable']) {
+  echo render($element['subject']);
+}
+else {
+  $e = [
+    '#theme' => 'form_element',
+    '#theme_wrappers' => [],
+    '#children' => '<div class="email-to-target-subject"><strong>' . check_plain($element['subject']['#default_value']) . '</strong></div>',
+  ] + $element['subject'];
+  echo render($e);
```
katzmo: I feel like the messageForm() function should already take care of that so `$element['subject']` is either the textfield or the plain string :thinking: Is that bad for backward compatibility?

katzmo (commented):
Really like the form template :heart_eyes:

campaignion_email_to_target/campaignion-email-to-target-email-message-form.tpl.php
torotil: No that would be possible as well … and I even did it like that at first, but preferred the theme-version then. It could be improved by adding another theme function for “textfield that renders as string with wrappers if disabled”.

katzmo: Hm, I think the  wrappers should be in the template, but it’s a bit much to build a renderable array there? 

Maybe something more like
```php
<?php ￼if (!$element['#editable']): ?>
  <p class="label email-to-target-subject-title"><?php print $element['subject']['#title'] ?></p>
  <p class="email-to-target-subject"><strong><?php print check_plain($element['subject']['#default_value']) ?></strong></p>
<?php endif; ?>
```
:thinking:

torotil: The reason for the renderable array is so that `theme_form_element()` can do its part.

torotil: … but yes. Maybe we can live without that. :thinking:

katzmo: Maybe we don’t even want that :wink: It would give us a label for an element that does not exist, no class to target it,…

torotil: Followed your suggestion wrt the markup. Thanks!

katzmo (approved):

PR #252, opened 2020-11-10
Template names & cleanups

torotil:
- Rename template files so that they allow follow the `theme-hook-name.tpl.php` naming convention.
- A bit more renaming in campaignion_bar.
- A bit of unused code removal.
- Update for travis.


campaignion_bar/theme/campaignion-bar.tpl.php
```diff
@@ -24,35 +24,4 @@
 
 <div id="ae-widepopups">
   <?php print theme('campaignion_bar_menu', array('widebar' => TRUE, 'links' => menu_tree_all_data('ae-menu'), 'wide' => array('New'))); ?>
```
katzmo, line 26: This is still `ae-menu`?

katzmo (commented):
Have the removed markup and styles never been used?

campaignion_bar/theme/campaignion-bar.tpl.php
torotil: Yes, I didn’t touch the menu name (or the styles).

torotil:
> Have the removed markup and styles never been used?

I have done a `ack hotkey-` over the code-base and that came up empty except for the code I’ve removed. The reason I looked more closely is because the template for this markup still has an advocacy engine link in it.

katzmo (approved):

PR #253, opened 2020-11-11
Make donation amount functionality generic

torotil:


campaignion_donation_amount/campaignion_donation_amount.info
```diff
@@ -0,0 +1,10 @@
+name = Campaignion donation amount total token
+description = Define a webform submission token that calculates the sum of two donation amounts.
```
katzmo: I think the description doesn’t quite cover what the module does?

campaignion_donation_amount/campaignion_donation_amount.variable.inc
```diff
@@ -0,0 +1,20 @@
...
+/**
+ * Implements hook_variable_info().
+ */
+function campaignion_donation_amount_variable_info($options) {
+  $v['campaignion_donation_amount_currency_symbol'] = [
+    'title' => t('Currency symbol'),
+    'description' => t('Currency symbol used for donation amounts.'),
+    'type' => 'string',
+    'default' => '€',
+    'localize' => FALSE,
```
katzmo: I agree that a currency symbol shouldn’t be translated, however practical experience shows it is. There is no other way to use different currency symbols for different languages that represent forms for different countries that use different currencies…

campaignion_giftaid/campaignion_giftaid.info
```diff
@@ -4,6 +4,7 @@ core = 7.x
 php = 7.0
 
 dependencies[] = campaignion:campaignion_donation
+dependencies[] = campaignion:campaignion_donation_amount
```
katzmo, line 6: Is the explicit dependency on campaignion_donation still needed now?

campaignion_donation_amount/campaignion_donation_amount.module
```diff
@@ -19,6 +19,10 @@ function campaignion_donation_amount_webform_component_render_alter(array &$elem
     $element['#element_validate'][] = 'campaignion_donation_amount_validate';
     // Always show the select_or_other textfield.
     $element['#attributes']['data-select-or-other-hide'] = 0;
+    if ($symbol = variable_get_value('campaignion_donation_amount_currency_symbol')) {
+      $element['#other_title'] = $symbol;
+      $element['#other_title_display'] = 'before';
+    }
```
katzmo, line 27: The amount component does not necessarily have to be of type `select_or_other` – do we want to add a "#other_title" to a text field or plain radios?

katzmo (changes requested):

campaignion_donation_amount/campaignion_donation_amount.info
torotil: True. Updated in the version that I just pushed.

campaignion_donation_amount/campaignion_donation_amount.variable.inc
torotil: I made it translatable.

campaignion_giftaid/campaignion_giftaid.info
torotil: Nope … and neither does campaignion_donation_amount really depend on campaignion_donation. I removed both dependencies.

campaignion_donation_amount/campaignion_donation_amount.module
torotil: Unused form-API `#properties` don’t hurt in any way. What’s better?

1. Set it unconditionally and have it ignored by other `#type`s
2. Add an additional check for `$element['#type'] == 'select_or_other'`?

katzmo: I'd prefer 2., then it's clearer in which cases we expect the property (and js attribute) to be used.

torotil: Updated with the latest commit.

campaignion_donation_amount/campaignion_donation_amount.info
katzmo: Better, thanks! For me "donation amount components" sound as if that was a component type though. Maybe it could be changed to something like "components used as donation amount"?

campaignion_donation_amount/campaignion_donation_amount.module
katzmo: Perfect, thanks!

campaignion_donation_amount/campaignion_donation_amount.info
torotil: They are “components” but not “component types”. I’d say `$component` is one specific configuration while `$component['type']` specifies which component plugin to use.

katzmo: I agree with the technical differentiation but that connotation does not get across in an English sentence :wink:

torotil: How does the difference between “components” and “component types” not come across in english?

katzmo (approved):

PR #254, opened 2020-11-26
Optin tracking

alex-mo:


campaignion_tracking/campaignion_tracking.info
```diff
@@ -2,3 +2,6 @@ name = Campaignion Tracking
 description = Common tracking functionality.
 core = 7.x
 package = Campaignion
+
+dependencies[] = campaignion:campaignion_action
+dependencies[] = little_helpers:little_helpers (>=2.0-alpha7)
```
torotil, line 8: A dependency on webform_paymethod_select:webform_paymethod_select is missing too.

campaignion_tracking/tests/TrackerTest.php
```diff
@@ -8,29 +11,31 @@
...
   }
 
   /**
-   * Delete test nodes.
+   * Tear down.
    */
   public function tearDown() {
-    node_delete($this->petition->nid);
     parent::tearDown();
   }
```
torotil: No need to keep this method around just for calling the parent method.

campaignion_tracking/tests/TrackerTest.php
```diff
@@ -110,17 +115,49 @@ public function testNodeContextDonation() {
...
     $nid = $submission->node->nid;
+
     campaignion_tracking_webform_redirect_alter($redirect, $submission);
+
     $this->assertEqual('t:t=s;w:nid=' . $nid . '&sid=1&title=Donation%20test%20%26%20stuff', $redirect->fragment);
 
     $redirect->fragment = 'something';
+
     campaignion_tracking_webform_redirect_alter($redirect, $submission);
+
```
torotil: I don’t think having an empty line after every line helps readability. You could group it into 3 parts: submission creation, creation of the redirect object + first assert, change of the redirect + second assert; each with an empty line it between.

campaignion_tracking/tests/TrackerTest.php
```diff
@@ -110,17 +115,49 @@ public function testNodeContextDonation() {
...
+    webform_component_defaults($component);
+    $this->petition->webform['components'][1] = $component;
+    $submission['sid'] = 1;
+    $submission['data'][1] = ['radios:opt-in'];
+    $submission = new Submission($this->petition, (object) $submission);
+    $submission->opt_in = new Values($submission);
+
+    campaignion_tracking_webform_redirect_alter($redirect, $submission);
+
+    $this->assertEqual('t:t=s;w:nid=1&sid=1&title=Test&optin[email]=opt-in', $redirect->fragment);
```
torotil, line 149: Same as above. Could create `$redirect` right before the alter-call.

torotil (changes requested):
Mostly comments except for the missing dependency.

campaignion_tracking/tests/TrackerTest.php
alex-mo: I wanted to split it by: setup/change → call → assert. This is indeed a bit weird if changing the redirect + assert is separated.

alex-mo: But I agree: putting the redirect object and the call closer together will improve readability

alex-mo: Better now?

campaignion_tracking/tests/TrackerTest.php
alex-mo: Removed

torotil (approved):
Awesome. Thanks!

campaignion_tracking/campaignion_tracking.module
```diff
@@ -126,9 +126,13 @@ function campaignion_tracking_webform_redirect_alter($redirect, Submission $subm
   $webformFragments[] = 'sid=' . $submission->sid;
   $webformFragments[] = 'title=' . rawurlencode($submission->node->title);
 
+  $channels = campaignion_opt_in_channel_info();
   // Add optin values for all defined optin fields
-  foreach ($submission->opt_in->values() as $optin) {
-    $webformFragments[] = 'optin[' . rawurlencode($optin['channel']) . ']=' . rawurlencode($optin['value']);
+  foreach ($channels as $channel => $title) {
+    $value = $submission->opt_in->canonicalValue($channel, TRUE);
+    if (!empty($value)) {
```
torotil: That’s the same as

```php
if ($value = $submission->opt_in->canonicalValue($channel, TRUE)) {
}
```

campaignion_tracking/campaignion_tracking.module
```diff
@@ -126,9 +126,13 @@ function campaignion_tracking_webform_redirect_alter($redirect, Submission $subm
   $webformFragments[] = 'sid=' . $submission->sid;
   $webformFragments[] = 'title=' . rawurlencode($submission->node->title);
 
+  $channels = campaignion_opt_in_channel_info();
   // Add optin values for all defined optin fields
-  foreach ($submission->opt_in->values() as $optin) {
-    $webformFragments[] = 'optin[' . rawurlencode($optin['channel']) . ']=' . rawurlencode($optin['value']);
+  foreach ($channels as $channel => $title) {
```
torotil: You could use `array_keys()` to avoid defining the unused variable `$title.`

Also you could loop over `array_keys(campaignionopt_in_channel_info())` unless you want to keep `$channels` for readability.

torotil (changes requested):
It’s possible that the test failure is not connected to the PR. This still blocks merging this PR.

Also two code-style suggestions.

torotil:
Update on the build failure. This is likely to be caused by Travis upgrading to xdebug 3.0.0 (last succesful build was 10 days ago with 2.9.8). Not sure yet how to deal with this. Maybe we need to upgrade phpunit.

campaignion_tracking/campaignion_tracking.module
alex-mo: thanks, keeping `$channels`

torotil (approved):

PR #256, opened 2020-12-11
[layout] Provide currently active layout as pseudo-global variable.

torotil:


torotil:
Closing in favor of #257 and #258.

PR #259, opened 2020-12-18
Option to skip campaignion_ux_webform_ajax_scroll

alex-mo:


campaignion_ux/js/campaignion_ux.js
```diff
@@ -19,6 +19,9 @@ $(document).ready(function() {
 // We want this behavior to execute last.
 Drupal.behaviors.campaignion_ux_webform_ajax_scroll = {};
 Drupal.behaviors.campaignion_ux_webform_ajax_scroll.attach = function(context) {
+  if (Drupal.settings.campaignion_ux.skip_scroll) {
+    return;
+  }
```
torotil, line 24: The reason this is needed is, that the functionality is now moved to campaignion_foundation. If we go down this path then I think we should remove this functionality from this module entirely and copy the behaviour to morelesszen instead. Is there any live themes that don’t have morelesszen (or campaignion_foundation) as base theme?

torotil (changes requested):

campaignion_ux/js/campaignion_ux.js
alex-mo: not entirely. the functionality is currently only re-implemented in a custom theme → we currently wanted only one specific theme to opt out of this code.

If we decide to remove this code we can just move all of it to morelesszen and remove the whole module...

alex-mo: I do not know if there are any live themes which do not have morelesszen as base theme...

torotil: So given the analysis of the legacy themes we’d need to add this to `ae_petition`, `morelesszen` ~~and `lesszen` (not yet entirely sure if it’s still used)~~.

alex-mo: Moved to themes

katzmo: Wouldn’t it be easier to just disable campaignion_ux for sites that don’t need it? :thinking:

alex-mo: Resolved in in-person discussion: Going with the quick solution now and coming back to `campaignion_ux` later...

torotil (approved):

campaignion_ux/js/campaignion_ux.js
```diff
@@ -19,6 +19,9 @@ $(document).ready(function() {
 // We want this behavior to execute last.
 Drupal.behaviors.campaignion_ux_webform_ajax_scroll = {};
 Drupal.behaviors.campaignion_ux_webform_ajax_scroll.attach = function(context) {
+  if (Drupal.settings.campaignion_ux.skip_scroll) {
```
katzmo: For most sites `Drupal.settings.campaignion_ux` is undefined, preventing AJAX from loading the next step. Please fix!

katzmo (changes requested):

katzmo (approved):

PR #262, opened 2021-01-12
e2t email address escaping

alex-mo:


campaignion_email_to_target/src/Message.php
```diff
@@ -31,11 +37,44 @@ class Message extends MessageTemplateInstance {
...
+      'fromAddress' => '[submission:values:email]',
+      'toName' => '[email-to-target:contact.title] [email-to-target:contact.first_name] [email-to-target:contact.last_name]',
+      'toAddress' => '[email-to-target:contact.email]',
       'display' => '[email-to-target:contact.display_name]',
     ];
     parent::__construct($data);
+
+    // Save as property for compatibility.
+    $this->to = $this->to();
+    $this->from = $this->from();
```
torotil: As far as I can see the only usage of these is in `src/Channel/Email.php` I think we can change it there instead of introding these for backwards compatibility.

campaignion_email_to_target/tests/MessageTest.php
```diff
@@ -105,4 +105,19 @@ public function testToArrayFromArray() {
...
+    $data = [
+      'fromName' => 'I have, a comma but am no "addr-list"',
+      'fromAddress' => 'from@example.com',
+      'toName' => ' to \ be " quoted',
+      'toAddress' => 'to@example.com',
+    ];
+    $m = new Message($data);
+    $this->assertEqual('"to \\\\ be \" quoted" <to@example.com>', $m->to);
+    $this->assertEqual('"I have, a comma but am no \"addr-list\"" <from@example.com>', $m->from);
+  }
```
torotil, line 121: Thanks for writing a test!

campaignion_email_to_target/src/Wizard/TargetStep.php
```diff
@@ -88,8 +88,11 @@ public function stepForm($form, &$form_state) {
     $settings['validations'] = [
       // Escape backslashes so JS won’t interpret them as escape sequence.
       'email' => '^[^\\s@]+@[^\\s@]+\\.[^\\s@]+$',
-      'first_name' => '\\S+',
-      'last_name' => '\\S+',
+      // First and last name are:
+      // - at least one non-whitespace char (which is not ,"@ too)
+      // - and no ,"@ chars in the whole string.
+      'first_name' => '^[^,"@]*[^,"@\\s][^,"@]*$',
+      'last_name' => '^[^,"@]*[^,"@\\s][^,"@]*$',
```
torotil: Would `'^\\s*([^,"@\\s]\\s*)+$'` be more readable? I would read that as whitespace followed by (at least one) whitespace separated words (with blacklisted chars).

torotil (changes requested):
Some tiny changes, some suggestions. Thanks for providing a test!

campaignion_email_to_target/src/Wizard/TargetStep.php
alex-mo: yes, probably more readable. (it's not a trivial Regexp anyways.) will change.

campaignion_email_to_target/src/Message.php
alex-mo: it's also used in `webform.php`

alex-mo: Changed.

torotil (approved):

PR #263, opened 2021-01-15
[layout] Implement theme default layouts.

torotil:


campaignion_layout/src/Lookup.php
```diff
@@ -70,10 +70,12 @@ public function getTheme() {
    * @return string|null
    *   The name of the first enabled layout of the currently active theme.
    */
-  public function getLayout() {
-    foreach ($this->iterateItems() as $item) {
-      if (($theme = $this->themes->getTheme($item['theme'])) && $theme->isActive() && ($layout = $theme->getLayout($item['layout']))) {
-        return $layout;
+  public function getLayout(string $active_theme_name) {
```
katzmo: The new argument still needs a doc string ;)

campaignion_layout/src/Lookup.php
```diff
@@ -70,10 +70,12 @@ public function getTheme() {
...
+    if ($theme = $this->themes->getTheme($active_theme_name)) {
+      foreach ($this->iterateItems() as $item) {
+        if ($active_theme_name == $item['theme'] && ($layout = $theme->getLayout($item['layout']))) {
```
katzmo: Just wondering, could this further be simplified by merging `getLayout()` and `getTheme()` into a `getItem()` function (as far as I can see they are both only called 1x)? Is there a chance that `$active_theme_name` is different to the one returned by `getTheme()`?

campaignion_layout/src/Lookup.php
torotil: I’ve removed the parameter again by moving the logic into a new `Theme::getLayoutFromItems()`.

campaignion_layout/src/Lookup.php
torotil: Hm, this is tricky as `Themes::getTheme()` and `Theme::getLayout()` are implemented on different classes and work with different data. How do you think about the new `Theme::getLayoutFromItems()`?

campaignion_layout/src/Themes.php
```diff
@@ -57,18 +57,44 @@ public function getTheme(string $theme_name = NULL) {
...
+   *
+   * The hook should be invoked on all enabled themes and all their base themes.
+   * Even disabled base themes should get their hook invoked as this matches how
+   * Drupal core treats themes.
+   */
+  protected function hookThemes() {
+    $enabled_themes = array_filter($this->allThemes(), function ($theme) {
+      return $theme->isEnabled() && $theme->hasFeature();
+    });
+    $hook_themes = $enabled_themes;
```
katzmo: This could just use the previous function: `$hook_themes = $this->enabledThemes()`?

campaignion_layout/tests/ThemesTest.php
```diff
@@ -55,23 +63,75 @@ public function testEnabledThemes() {
...
+    $theme->expects(isset($layouts) ? $this->once() : $this->never())
+      ->method('invokeLayoutHook')->willReturn($layouts);
+    return $theme;
+  }
+
+  /**
+   * Test getting the declared layouts with a disabled base theme.
+   */
+  public function testDeclaredLayoutsDisabledBase() {
+    $theme['foo_base'] = $this->mockTheme([], TRUE, TRUE, []);
```
katzmo: The test name says "disabled base" but with `TRUE` as second argument it is not actually disabled, is it? It just doesn’t declare any layouts?

campaignion_layout/src/Themes.php
torotil: Not quite. `$enabled_themes` doesn’t (and shouldn’t) check whether the feature toggle is enabled in the settings. So this is not the same as `$this->enabledThemes()`.

torotil: What about renaming the variable to `$layout_themes`?

campaignion_layout/tests/ThemesTest.php
torotil: Fixed, thanks. It still checks whether the hook was invoked on the base theme though.

campaignion_layout/src/Lookup.php
katzmo: Oh no, I was only referring to the `Lookup` class!

```php
  // E.g. `Lookup::getTheme()` could return the item instead of the theme
  public function getItem() {
    foreach ($this->iterateItems() as $item) {
      if (($theme = $this->themes->getTheme($item['theme'])) && $theme->hasFeatureEnabled()) {
        return $item;
      }
    }
  }

// Then somewhere else
$theme = $lookup->getItem()['theme'];
$layout = $lookup->getItem()['layout'];
```

But `Theme::getLayoutFromItems()` is also an improvement to the previous code :)

torotil: That would still not work:

- For the theme we choose the first item with a feature-enabled theme.
- For the layout we choose the first item of the currently active theme which contains a layout that’s enabled for this theme.

Your code would use the same criteria for both to choose the item.

campaignion_layout/src/Themes.php
katzmo: Ah, yes, `hasFeature()` vs. `hasFeatureEnabled()`, that’s easy to overlook…

I like that the variable matches the function name, but `layoutThemes()` could be better in providing information what those themes are used *for* instead of *where*?

campaignion_layout/src/Lookup.php
katzmo: That’s was what I was wondering in the first comment, *is* that a difference? Shouldn’t the first selected theme be the theme that is used?

torotil: @katzmo If we assume nothing else interferes with theme selection: yes. Even then it’s still not as simple:

```php
$items[] = ['theme' => 'a', 'layout' => 'not available'];  // Choose theme a because of this.
$items[] = ['theme' => 'b', 'layout' => 'available layout 1'];  // Don’t use this because 'b' is not the active theme.
$items[] = ['theme' => 'a', 'layout' => 'available layout 2'];  // Use this layout because it’s the first one for the active theme.
```

This should choose the third item, not the first. (Assuming both themes have the feature enabled.)

campaignion_layout/src/Themes.php
torotil: Another thought to make this simpler. The main complication of working out the list of themes upfront is caused by the assumption that each theme should be invoked only once (mainly because `drupal_array_merge_deep($array, $array) == $array` doesn’t hold for all inputs). We could instead do the following:

1. `hook_campaignion_layout_info()` gets the base theme’s return value (or an empty array) as parameter.
2. The themes are responsible for keeping or discarding their base theme’s layouts.
3. If a theme doesn’t implement the hook then the base theme’s return value is returned.

Disadvantage: This would mean layouts could appear/disappear based on whether the base theme is enabled (if the child theme doesn’t pass them through). ^^

katzmo: I like the parameter idea, but we'd need to call the hook of disabled base themes…which probably isn’t possible?

torotil: Surely possible … and exactly what’s happening with `Themes::hookThemes()` right now. The difference between `$enabled_themes` and `$hook_themes` is exactly the disabled base themes.

katzmo (approved):

PR #273, opened 2021-04-07
Switchable content/form order for layouts

alex-mo:


campaignion_layout/campaignion_layout.install
```diff
@@ -20,10 +20,49 @@ function campaignion_layout_field_schema($field) {
...
+      );
+      foreach ($table_prefixes as $table_prefix) {
+      }
+      $field_name = $field['field_name'];
+      $table1 = _field_sql_storage_tablename($field);
+      $table2 = _field_sql_storage_revision_tablename($field);
+      $column = $field_name . '_order';
+
+      db_add_field($table1, $column, $spec);
+      db_add_field($table2, $column, $spec);
```
torotil: This looks like two versions of the same code.

campaignion_layout/campaignion_layout.module
```diff
@@ -79,6 +79,11 @@ function campaignion_layout_page_build(&$page, $node = NULL) {
         $plugin->execute($layout['name']);
       }
     }
+    $page['#order_enabled'] = $layout['order_enabled'] ?? FALSE;
+    $page['#order'] = 'above-main';
```
torotil: This is reading “page order is above main” but it should actually be more “form above main”, right?

Also: Could we squeeze this into one variable that is either an order or `FALSE`/`NULL`?

campaignion_layout/src/Lookup.php
```diff
@@ -71,4 +71,19 @@ public function getLayout() {
     return $this->themes->getTheme()->getLayoutFromItems($this->iterateItems());
   }
 
+  /**
+   * Get the order for the form for the currently active layout.
+   *
+   * @return string|null
+   *   The machine name of the ordering.
+   */
+  public function getOrder() {
```
torotil: I don’t think this makes any sense wrt multiple items. This might choose one item for the layout and then another one for the order. Maybe we should switch to returning the item instead and then reading all its options?

torotil (changes requested):

torotil:
Also this makes the tests fail at the moment.

campaignion_layout/campaignion_layout.module
alex-mo: I do not cling to this specific naming, I thought `above-main` is good enough. but sure, we can extend to `form-above-main` or use `above-content`, which might be more accurate. whichever you like most (I think a fresh look on the naming here is helpful indeed :innocent:)...

alex-mo: I choose a string because they are easier extensible with additional values than a booleans, which might become a issue with this field. yes, if we are confident enough, that this can stay a boolean, OK with me.

campaignion_layout/campaignion_layout.install
alex-mo: `$table1 !== $table2`. or what do you mean with "same code"?

alex-mo: ah, the code above. yes. that's a leftover? :innocent:

campaignion_layout/src/Lookup.php
alex-mo: Different items how? It's checking for the layout name and if the layout enables the ordering...

campaignion_layout/campaignion_layout.module
torotil: Naming: Yep, something that doesn’t suggest that the page is rendered above main (which doesn’t make sense) would be better.

For compressing this into one variable I was suggesting something more like `$page['#order']` being either `FALSE` or maybe even `'default'` as long as `$layout['order_enabled']` is false. So only setting `$page['#order']` to some value when it’s also enabled. This way the theme doesn’t have to implement the toggle logic.

campaignion_layout/src/Lookup.php
torotil: > Different items how?

Can you guarantee that there is only one field item with the said layout name?

Anyway getting the item only once is prettier I think.

campaignion_layout/campaignion_layout.module
alex-mo: I would take up the suggestion from @katzmo to call it `reversed`.

campaignion_layout/src/Lookup.php
alex-mo: Sorry. I do not get the items stuff here. Why is it different to `getTheme()` or `getLayoutFromItems()` which also return the first matching item. I thought this is how the layout stuff is allowed to work... :thinking:

torotil: `getTheme()` and `getLayoutFromItems()` run at different times. It could even be that one runs in request in which the other isn’t called at all.

`getLayoutFromItems()` and `getOrder()` on the other hand would both run during page preprocessing.

campaignion_layout/campaignion_layout.module
torotil: Ok, works as well but doesn’t have the advantage of being extensible. Fine for me though ;-)

campaignion_layout/src/Lookup.php
torotil: Implemented my own suggestion.

campaignion_layout/campaignion_layout.module
torotil: Implemented my own suggestion but left `#reversed` as-is.

katzmo (approved):

torotil (approved):

PR #279, opened 2021-06-11
Cookie bar cleanup

torotil:


campaignion_google_analytics/campaignion_google_analytics.module
```diff
@@ -228,3 +228,21 @@ function campaignion_google_analytics_page_build(&$page) {
...
+ */
+function campaignion_google_analytics_campaignion_tracking_snippets() {
+  $snippets = [];
+  if ($ga_id = variable_get_value('campaignion_google_analytics_id')) {
+    $snippets['ga'] = <<<SNIPPET
+(function(i,s,o,g,r,a,m){i['GoogleAnalyticsObject']=r;i[r]=i[r]||function(){
+(i[r].q=i[r].q||[]).push(arguments)},i[r].l=1*new Date();a=s.createElement(o),
+m=s.getElementsByTagName(o)[0];a.async=1;a.src=g;m.parentNode.insertBefore(a,m)
+})(window,document,'script','https://www.google-analytics.com/analytics.js','ga');
+ga('create', '$ga_id', 'auto');
```
katzmo: `$ga_id` should probably not be quoted :wink:

torotil: The quotes are needed in this version (otherwise it would be interpreted as a JS variable, or worse executed as code), but it’s probably much better to JSON encode it instead than to hard-code the quotes :thinking:

katzmo: Right, the heredoc is so long I forgot it's there :see_no_evil:

Not sure JSON encoding is worth the trouble? Maybe putting it in curly braces would make the variable easier to recognize? :thinking:

katzmo (approved):

campaignion_google_analytics/campaignion_google_analytics.module
torotil: I did both :-)

campaignion_tracking/campaignion_tracking.module
```diff
@@ -184,13 +184,13 @@ function campaignion_tracking_webform_redirect_alter($redirect, Submission $subm
...
 function _campaignion_tracking_snippets() {
   $snippets = module_invoke_all('campaignion_tracking_snippets');
-  drupal_alter('campaignion_tracking_snippets', $snippets);
   foreach ($snippets as &$snippet) {
     if (!is_array($snippet)) {
       $snippet = ['js' => $snippet];
     }
     $snippet += ['event' => 'tracking-allowed'];
   }
+  drupal_alter('campaignion_tracking_snippets', $snippets);
```
katzmo, line 193: Is that a benefit? The difference I see is that other modules now need to provide their additional snippets as array, including an event name, while before just the snippet would have been enough and the event name optional. The old version seemed safer to me. Am I overlooking anything else here? :thinking:

torotil: Jep. The original hook is called before the normalization. So if you just *provide* snippets you can rely on the normalization. If on the other hand you want to manipulate snippets defined elsewhere you can rely on a canonical format, because the alter hook is called after the normalization.

katzmo (approved):

campaignion_tracking/campaignion_tracking.module
torotil: I mean:

* If you want to define snippets: Use `hook_campaignion_tracking_snippets()` and rely on the normalization.
* If you want to change snippets: Use `hook_campaignion_tracking_snippets_alter()` and rely on getting the normalized snippets as input for altering.

katzmo: Makes sense :+1:

PR #281, opened 2021-07-09
Add module for internal node admin titles

katzmo:


campaignion_node_admin_title/campaignion_node_admin_title.features.field_base.inc
```diff
@@ -0,0 +1,37 @@
...
+ */
+
+/**
+ * Implements hook_field_default_field_bases().
+ */
+function campaignion_node_admin_title_field_default_field_bases() {
+  $field_bases = array();
+
+  // Exported field_base: 'field_admin_title'.
+  $field_bases['field_admin_title'] = array(
```
torotil, line 15: I usually rename pre-defined fields to something that doesn’t have the `field_` prefix (`eg. admin_title` in this case). This way I know that they are provided by some module and probably have functionality attached.

torotil (changes requested):

campaignion_node_admin_title/campaignion_node_admin_title.features.field_base.inc
katzmo: I tried `admin_title` but it didn’t work (field never showed up) :shrug: I can give it another try or try a different name if you find it important…

torotil: I think `admin_title` doesn’t work because it’s already in use in `$node->admin_title`, but not super-important to me.

torotil (approved):

PR #284, opened 2021-07-29
[facebook_pixel] fix fbq dummy

katzmo:
Modify snippet so it still loads fbevents.js when a dummy exists.


torotil (approved):
Awesome debugging work. Sorry, that this slipped through.

torotil:
Looking at this again I don’t think we even need to take care of creating the dummy in the snippet itself. I think it is (or should be) guaranteed that the `fb-events.js` is loaded before the snippet is executed.

PR #285, opened 2021-08-31
Switched from load to DomContentLoaded event

jamie-moreonion:
I've updated this widget as part of a solution to the MO website's 'Contact Us' form loading very slowly... it turns out that the time difference between the `DOMContentLoaded` event and the `load` event is in the order of 600-1000ms, which is almost certainly due to the recaptcha script. I see that Alex was the original developer for this, but as he's not here, could I ask for a little verification that this is not going to break anything I don't know about? :)

### DIFF
--- a/campaignion_webform_widget/js/widget.js
+++ b/campaignion_webform_widget/js/widget.js
@@ -49,18 +49,14 @@ function gaLinkerHandler() {
   }
 }
 
-if (window.addEventListener) {
-  window.addEventListener("DOMContentLoaded", function() {
-    gaLinkerHandler();
-    messageParent();
-    window.addEventListener("DOMSubtreeModified", messageParent, true);
-  });
-}
-else {
-  window.onload = function() {
-    messageParent();
-    window.attachEvent("onDOMSubtreeModified", messageParent);
-  }
+window.addEventListener("DOMContentLoaded", function() {
+  gaLinkerHandler();
+  messageParent();
+});
+
+window.onload = function() {
+  messageParent();
+  window.addEventListener("DOMSubtreeModified", messageParent, true);
 }
 
 window.onresize = function() {
--- a/campaignion_webform_widget/js/widget.js
+++ b/campaignion_webform_widget/js/widget.js
@@ -49,18 +49,20 @@ function gaLinkerHandler() {
   }
 }
 
-function registerListeners() {
-  if (window.addEventListener) {
-    window.addEventListener("DOMSubtreeModified", function(){messageParent();}, true);
-    document.addEventListener("DOMContentLoaded", gaLinkerHandler);
-  } else if (window.attachEvent) {
-    window.attachEvent("onDOMSubtreeModified", function(){messageParent();});
+if (window.addEventListener) {
+  window.addEventListener("DOMContentLoaded", function() {
+    gaLinkerHandler();
+    messageParent();
+    window.addEventListener("DOMSubtreeModified", messageParent, true);
+  });
+}
+else {
+  window.onload = function() {
+    messageParent();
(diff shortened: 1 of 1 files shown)


campaignion_webform_widget/js/widget.js
```diff
@@ -49,18 +49,20 @@ function gaLinkerHandler() {
...
-    window.addEventListener("DOMSubtreeModified", function(){messageParent();}, true);
-    document.addEventListener("DOMContentLoaded", gaLinkerHandler);
-  } else if (window.attachEvent) {
-    window.attachEvent("onDOMSubtreeModified", function(){messageParent();});
-  }
+if (window.addEventListener) {
+  window.addEventListener("DOMContentLoaded", function() {
+    gaLinkerHandler();
+    messageParent();
+    window.addEventListener("DOMSubtreeModified", messageParent, true);
```
katzmo: Why register the "DOMSubtreeModified" handler within the "DOMContentLoaded" handler?

campaignion_webform_widget/js/widget.js
```diff
@@ -49,18 +49,20 @@ function gaLinkerHandler() {
...
+  });
 }
-window.onload = function() {
-  messageParent();
-  registerListeners();
+else {
+  window.onload = function() {
```
katzmo: I agree we can get rid of `window.attachEvent`, however, I don't see a reason to keep a *fallback* on `window.onload` instead? I rather think we need to *always* call `messageParent()` again on load, in case there were some images loaded or scripts that add content and the document size changed.

katzmo (changes requested):
The module is used on various sites, so there is definitely potential for breaking a widget without noticing ;)

campaignion_webform_widget/js/widget.js
jamie-moreonion: I placed it within the `DOMContentLoaded` handler for two reasons:
- The old `registerListeners` function that previously contained it was originally activated on `window.load`; having chosen instead to use the `DOMContentLoaded` event to represent the page load, it seemed to be the most equivalent position for it in the new structure.
- I am not 100% sure if this is the case, but it occurred to me that setting up a `DOMSubtreeModified` listener before the DOM had finished loading might cause a deluge of pointless events... so I kept it within the `DOMContentLoaded` handler in an effort to prevent this (see also comment below).

campaignion_webform_widget/js/widget.js
jamie-moreonion: I think this is probably what the `DOMSubtreeModified` listener is supposed to be doing... I'd like to be more certain about this, but documentation is sparse as it belongs to a deprecated API and I'm having issues testing it via the `campaignion_standalone` box. As for the `window.onload` fallback, I only included it to represent a path for clients that don't have `window.addEventListener` - i.e. IE < 9 - but I'd be comfortable about removing it.

katzmo: We don't support IE < 9 more.

The "DOMSubtreeModified" listener informs the parent that the iframe needs to be resized when the DOM is modified but not when an image (without a specified height) that was always in the DOM has finished loading… (just an example why load is more reliable regarding the actual document length, widgets don't usually load big images)

jamie-moreonion: Yes, I agree that the reflow on image loading could be an issue... thank you for pointing it out. As per our conversation earlier, I'd be quite happy not to go ahead with this change in the interest of not breaking other uses of the widget, especially as this code is up for replacement anyway - but if we did want to do anything, perhaps your idea of using both the `DOMContentLoaded` and `load` events in series to resize might give us the best of both worlds? I've updated the branch with that for you to take a look, plus I've removed the IE9 support.

katzmo (approved):

PR #287, opened 2021-09-06
Create activities asynchronously

torotil:
Synchronous creation of activities comes from a time when recent supporter features were still based on activities. This is no longer the case since quite a few years.

Removing the synchronous processing removes a lot of complexity:

- Contacts no longer need to be created in two phases.
- Better performance and less sources of error during form submission because less processing is done during the form submission request.

----

The newsletter subscription activities are written in a way that they need the synchronously created contacts and activities. They are also made obsolete by the opt-in activities to some degree. They are not actually *used* for anything. That’s why it seems much easier to just remove them.


campaignion_action/campaignion_action.module
```diff
@@ -79,13 +79,20 @@ function campaignion_action_module_implements_alter(&$impl, $hook) {
     }
     $impl = $new_impl;
   }
-  if ($hook == 'webform_redirect_alter') {
+  $module = 'campaignion_action';
+  $move_to_front = function () use (&$impl, $module) {
```
katzmo: What’s `$module` for, it doesn’t seem to be used anywhere?

campaignion_activity/campaignion_activity.module
```diff
@@ -16,22 +19,37 @@ function campaignion_activity_views_api() {
...
+  }
+
+  $activity = WebformSubmission::fromSubmission($submission, $contact);
+  $activity->contact_id = $contact->contact_id;
+  $activity->confirmed = $activity->confirmed ?? ($submission->confirmed ? $when : NULL);
+
+  if ($payments = $submission->payments ?? NULL) {
+    foreach ($payments as $payment) {
+      if (payment_status_is_or_has_ancestor($payment->getStatus()->status, PAYMENT_STATUS_SUCCESS)) {
+        $activity = new WebformPayment($activity);
```
katzmo, line 38: Don’t we want a WebformSubmission and a WebformPayment activity?

torotil: This doesn’t create another activity it extends the existing one. So there will be one activity with both entries in `campaignion_activity_webform` and `campaignion_activity_payment`. I wouldn’t implement it like this again, but that’s what the behaviour was before this PR.

campaignion_action/campaignion_action.module
torotil: I’ve removed it in a new commit.

campaignion_activity/campaignion_activity.module
katzmo: OK

katzmo (approved):

PR #293, opened 2021-11-16
CI: Test on PHP 7.4

torotil:


campaignion_wizard/templates/campaignion-wizard-thank-summary.tpl.php
```diff
@@ -29,10 +31,12 @@
 <?php
 if ($after_confirm) {
   if ($after_confirm['type'] == 'node') {
-    echo t('After clicking on the confirmation link your supporters will see the page <a href="@page_href">@page_title</a>.', [
-      '@page_href' => url("node/{$after_confirm['node_reference_nid']}"),
-      '@page_title' => $after_confirm['node']->title,
-    ]);
+    if ($after_submit['node']) {
```
katzmo: Should this be `$after_confirm['node']`?

torotil: Thanks for spotting this. I think I was relying too much on my colored variables that don’t work for `.tpl.php` files. I’ve pushed a fixed version of the commit.

katzmo (approved):

PR #298, opened 2022-01-28
Restructure expiry functions

katomat:
Goal:
Have a function for other scripts to anonymise individual contacts.
Add a little_helpers service to anonymise a contact.
Expiry Cron jobs should call it instead of providing this functionality themselves.

### DIFF
--- a/campaignion_expiry/src/Anonymizer.php
+++ b/campaignion_expiry/src/Anonymizer.php
@@ -1,0 +1,88 @@
+<?php
+
+namespace Drupal\campaignion_expiry;
+
+use Drupal\campaignion\Contact;
+
+/**
+ * Anonymization service for redhen contacts.
+ */
+class Anonymizer {
+
+  /**
+   * Anonymize a single contact.
+   *
+   * @param bool $keep_mp_fields
+   *   If TRUE the MP data is copied to the anonymized contacts, otherwise it is
+   *   deleted.
+   */
+  public function __construct($contact, bool $keep_mp_fields = FALSE) {
+    $this->contact = $contact;
+    $this->keepMpFields = $keep_mp_fields;
+    $this->entityController = entity_get_controller('redhen_contact');
+  }
+
+  /**
+   * Create a new anonymous contact and copy over some data we want to keep.
+   *
+   * @param \Drupal\campaignion\Contact $contact
+   *   Contact that is being anonymized.
+   */
+  public function anonymize(Contact $contact) {
+    $new_contact = new Contact();
+    $new_contact->contact_id = $contact->contact_id;
+    $new_contact->redhen_state = REDHEN_STATE_ARCHIVED;
+    $new_contact->setEmail("{$contact->contact_id}@deleted");
+
+    $new_contact->first_name = 'Anonymous';
+    $new_contact->last_name = $contact->contact_id;
+
+    // Copy first country of $contact to $new_contact.
+    foreach ($contact->wrap()->field_address->value() as $item) {
+      if (!empty($item['country'])) {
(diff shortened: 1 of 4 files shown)


katomat:
[Campaignion: Expiry](https://trello.com/c/EOHDgG5v/1316-campaignion-expiry)

campaignion_expiry/src/Anonymizer.php
```diff
@@ -0,0 +1,90 @@
...
+/**
+ * Anonymization service for redhen contacts.
+ */
+class Anonymizer {
+
+  /**
+   * Anonymize a single contact.
+   *
+   * @param \Drupal\campaignion\Contact $contact
+   *   Contact that is being anonymized.
```
torotil: The `Anonymizer` gets the contact passed as argument to the method call not to the constructor. That would also match the service definition.

```suggestion
```

campaignion_expiry/tests/AnonymizeContactTest.php
```diff
@@ -0,0 +1,44 @@
...
+use Drupal\campaignion\Contact;
+
+/**
+ * Test anonymizing a contact.
+ */
+class AnonymizerTest extends DrupalUnitTestCase {
+
+  /**
+   * Set up.
+   */
```
torotil: A more informative comment would be nice (also below at `tearDown()`.

campaignion_expiry/tests/AnonymizeContactTest.php
```diff
@@ -0,0 +1,44 @@
...
+  /**
+   * Test anonymizing our contact.
+   */
+  public function testContactAnonymization() {
+    $anonymizer = new Anonymizer(false);
+    $anonymizer->anonymize($this->contact, false);
+    $contact = entity_load_single('redhen_contact', $this->contact->contact_id);
+    $this->assertNotEmpty($contact);
+    $this->assertEqual($contact->first_name, "Anonymous");
+  }
```
torotil, line 52: We could/should also test `Anonymizer::deleteOldRevisions()` here.

katomat:
I added more comments and added testing for the deleteOldRevisions functions.

campaignion_expiry/tests/AnonymizeContactTest.php
```diff
@@ -0,0 +1,57 @@
+<?php
+
+namespace Drupal\campaignion_expiry;
+
+use Drupal\campaignion\CRM\Import\Source\ArraySource;
+use Drupal\campaignion\ContactTypeManager;
+use Drupal\campaignion_expiry\Anonymizer;
+use Upal\DrupalUnitTestCase;
+use Drupal\campaignion\Contact;
```
torotil: Could you sort the use statements alphabetically?

campaignion_expiry/tests/AnonymizeContactTest.php
```diff
@@ -0,0 +1,57 @@
...
+
+/**
+ * Test anonymizing a contact.
+ */
+class AnonymizerTest extends DrupalUnitTestCase {
+
+  /**
+   * Set up.
+   *
```
torotil: No need to keep the worthless part of the comment. ;-)

torotil (changes requested):
… otherwise good enough for me :-)

katomat:
@torotil Wouldn't the tearDown actually require to not only delete the $this->contact, but also the actual anonymised redhen_contact?

torotil:
> @torotil Wouldn't the tearDown actually require to not only delete the $this->contact, but also the actual anonymised redhen_contact?

They are the same — as in having the same `$contact->contact_id` that’s what anonymization means after all.

torotil (approved):

PR #301, opened 2022-03-03
Implement loqate UK account validation

katomat:


campaignion_loqate/package/.browserslistrc
```diff
@@ -0,0 +1,7 @@
+last 1 version
+> 1%
+maintained node versions
+not dead
+not ie > 0
+not ie_mob > 0
+not android > 0
```
katzmo: The difference of the list compared to `defaults, maintained node versions` is (currently) excluding android 98, firefox 91, firefox 78, ie 11, ios_saf 14.0-14.4, ios_saf 12.2-12.5, safari 15.1, samsung 15.0 – is that on purpose? I can see why one might want to explicitly exclude IE 11, but why the android browser? Mobile IE is dead so no need to exclude it extra.

campaignion_loqate/package/src/client.js
```diff
@@ -0,0 +1,81 @@
...
+var $ = jQuery
+
+/**
+ * API client for the loqate JSON endpoints.
+ */
+class Client {
+  constructor (settings) {
+    this.settings = settings
+  }
+  validateSortCode (sortCode) {
```
katzmo, line 12: Why do we need an extra validation for sort codes, independent of the account? `validateAccount` validates the sort code too, is there a reason to enforce a valid sort code before checking sort code + account number?

campaignion_loqate/package/src/validator.js
```diff
@@ -0,0 +1,85 @@
...
+    this.disabledButtons = $()
+  }
+  bind (wrapper) {
+    this.inputHolder = (new Input($('[name$="[holder]"]', wrapper), Drupal.t('Please enter the account holder’s name.'))).bind()
+    this.inputAccount = (new Input($('[name$="[account]"]', wrapper), Drupal.t('Please enter a valid account number.'))).bind()
+    this.inputSortCode = (new Input($('[name$="[bank_code]"]', wrapper), Drupal.t('Please enter a valid sort code.'), [this.inputAccount])).bind()
+    let validate = debounce(() => {
+      this.validate()
+    }, this.settings.validateTimeout)
+    $(wrapper).on('change keyup', validate)
```
katzmo: This ignores the clientside validation setting of *when* fields should be validated. But I guess this module doesn’t have to be perfect :wink:

campaignion_loqate/package/src/input.js
```diff
@@ -0,0 +1,76 @@
...
+  }
+  unmarkIfChanged () {
+    if (this.hasChanged()) {
+      this.unmark()
+      this.dependent.forEach(function (other) {
+        other.unmark()
+      })
+    }
+    else {
+      this.mark()
```
katzmo: Why is this needed when nothing has changed?

campaignion_loqate/package/src/input.js
```diff
@@ -0,0 +1,76 @@
...
+    this.$input.on('change keyup', () => {
+      this.unmarkIfChanged()
+    })
+    return this
+  }
+  unmark () {
+    const validator = Drupal.myClientsideValidation.validators[this.formId]
+    const $wrapper = $('#clientsidevalidation-' + this.formId + '-errors')
+    const errors = validator.errorsFor(this.$input.get(0))
+    validator.addWrapper(errors).remove()
```
katzmo: Doesn’t this need something like `validator.hideErrors()`? :thinking:  
What does `.addWrapper().remove()` do?

campaignion_loqate/package/src/client.js
torotil: Possible reasons are:

- We want to validate the sort code before the account number is entered. I could work around this by supplying a dummy value for the account in this case, but it’s not pretty.
- It’s harder to assign the error to a field if the API-call is used for both.

campaignion_loqate/package/src/validator.js
torotil: Where would I find this setting? In general think the module was more “perfect“ before the integration with clientside_validation. I would settle for a good compromise between integration / simplicity / UX in this case.

campaignion_loqate/package/src/input.js
torotil: The code is mainly transferred from [here](https://github.com/moreonion/stripe_payment/blob/7.x-2.x/package/src/method-element.js#L123) and the [clientSideValidationErrorHandler](https://github.com/moreonion/stripe_payment/blob/7.x-2.x/package/src/method-element.js#L235) I don’t know if / when the code of [resetValidation](https://github.com/moreonion/stripe_payment/blob/7.x-2.x/package/src/method-element.js#L306) should be called.

Both are good questions, but I don’t know the stripe code close enough to answer them. Do you remember what these lines do and when they are needed?

campaignion_loqate/package/src/input.js
torotil: I couldn’t find a documented reason by looking at the history of foeuk_loqate.

When I remove it I can reproduce the following bug:

1. Enter invalid sort code → error displayed
2. Focus the account field
3. Focus the sort code field
4. Focus the account field → error on the sort-code removed

EDIT: It seems that that’s due to something clientside_validation does on its own.

campaignion_loqate/package/src/validator.js
katzmo: I think the "right" why to do this is registering a custom validator function and let clientside validation call it when it sees fit.

See example on https://git.drupalcode.org/project/clientside_validation/-/blob/7.x-1.47/clientside_validation.api.js#L25 (this expects an equivalent server side validation, without I guess a rule needs to be added manually to jquery validate for the field to use it?)

campaignion_loqate/package/src/input.js
katzmo: Have looked it up: if elements have a wrapper, `addWrapper(elements)` adds the wrappers to elements so `hide()` will hide the errors *and* their wrappers. Sounds like the right choice here.

`hideErrors()` on the other hand will hide all errors on the form. I think our `resetValidation()` function was needed when switching between payment forms so you won’t get e.g. Stripe validation errors after switching to a DD form? :thinking:

campaignion_loqate/package/src/validator.js
torotil: I tried to do that. I already had a custom remote validation that works with their API and jQuery validate (pre-promises async code).

The problem I got stuck at was on the PHP side: actually making clientside_validate *configure* the validation for these fields based on something in the form. I tried to do this based on form-API #element_validate functions, but I couldn’t make it work as my implementation of [`hook_clientside_validation_rule_alter()`](https://git.drupalcode.org/project/clientside_validation/-/blob/7.x-1.x/clientside_validation.api.php#L71) was never called. webforms seem to be processed entirely differently than normal form-API forms.

campaignion_loqate/package/.browserslistrc
torotil: I think the reason for this particular list was that it made the bundle much smaller to exclude all these browsers. This was 2 years ago.

torotil: It doesn’t seem to make a big difference today. So I can just use defaults / remove `.broswerlistrc` is that the preferred way of doing it?

katzmo: I think the file is needed, but it can just include `defaults`.

campaignion_loqate/package/src/validator.js
katzmo: Well, unless we want to validate server side too, I think this could be done in JS only?

campaignion_loqate/package/.browserslistrc
torotil: The output is the same for either having a file with `defaults` or having none. I removed the file.

campaignion_loqate/package/src/validator.js
torotil: Ok, how? How would I register the validators for the fields (without breaking other clientside validations)?

campaignion_loqate/package/.browserslistrc
katzmo: Maybe that's just parcel then.

campaignion_loqate/package/src/validator.js
katzmo: Something along the lines of this maybe? (haven’t actually tried it)

```js
$(document).bind('clientsideValidationAddCustomRules', function(event){
    // see https://jqueryvalidation.org/jQuery.validator.addMethod/
    jQuery.validator.addMethod("myCustomMethod", function(value, element, param) {
        // do the validation
    }, jQuery.format('Error message'));

    // not sure when this needs to happen exactly
    // see https://jqueryvalidation.org/rules/
    jQuery(field).rules("add", {"myCustomMethod": true})
});

```

torotil:
```javascript
    // not sure when this needs to happen exactly
    // see https://jqueryvalidation.org/rules/
    jQuery(field).rules("add", {"myCustomMethod": true})
 ```
 
 Your “not sure …” is exactly what I was asking for. The rest was clear. I think I found a solution by adding the appropriate rule-settings in a `#after_build` callback — this allows me to still ignore all the PHP stuff clientside_validation does.

Comment (katzmo) [reply campaignion_loqate/package/src/validator.js]: Not in the first version of your question…

Maybe there is a "too early" but there doesn’t seem to be a "too late" – when I add a rule in the browser console on a page it just works.

File: campaignion_loqate/package/src/client.js (hunk 2 above)
Comment (torotil) [reply campaignion_loqate/package/src/client.js:12]: Ok, and now I’ve tested that too: Although the API could in theory return the appropriate error for both fields (it’s a list of errors) it only returns one at a time. So as long as the account number is not valid I don’t get an indication of the sort code’s validity. So I guess using both endpoints is the only option.

Comment (katzmo) [reply campaignion_loqate/package/src/client.js:12]: Ok, thanks. But that does sound like we could check the account number without having a valid sort code?

Comment (torotil) [reply campaignion_loqate/package/src/client.js:12]: > But that does sound like we could check the account number without having a valid sort code?

As the sort code is still a required parameter of the API-call for checking the account I still don’t think so. I think what we’ll see is:

- sortCode="", accountNumber="" → account required
- sortCode="", accountNumber="not the right number of digits" → account invalid
- sortCode="", accountNumber="right number of digits but invalid for the sort code" → sort code missing
- sortCode="invalid", accountNumber="right number of digits but invalid for the sort code" → sort code invalid
- sortCode="valid", accountNumber="right number of digits but invalid for the sort code" → account invalid
- …

Comment (katzmo) [reply campaignion_loqate/package/src/client.js:12]: Evidently makes no sense to validate one field without the other :shrug:  
But I guess the postcode-API is the best possible approach to attempt it anyway.

Comment (torotil) [reply campaignion_loqate/package/src/client.js:12]: It makes sense to validate the sort code without the account, but not the other way around. :shrug:

File: campaignion_loqate/campaignion_loqate.info [hunk 6]
```diff
@@ -5,3 +5,5 @@ core = 7.x
 dependencies[] = clientside_validation:clientside_validation
 dependencies[] = manual_direct_debit_uk
 dependencies[] = variable
+
+scripts[] = campaignion_loqate.min.js
```
katzmo, line 9: When the js is added in the info file, it doesn’t need `$form['#attached']['js'][] = drupal_get_path('module', 'campaignion_loqate') . '/campaignion_loqate.min.js';` anymore, does it?

campaignion_loqate/campaignion_loqate.module
```diff
@@ -0,0 +1,42 @@
...
+ * Implements hook_payment_forms_payment_form_alter().
+ */
+function campaignion_loqate_payment_forms_payment_form_alter(&$form, \Payment $payment) {
+  if (!($payment->method->controller instanceof AccountDataController)) {
+    return;
+  }
+  if (!($key = variable_get_value('campaignion_loqate_api_key'))) {
+    return;
+  }
+  $form['#attributes']['class'][] = 'loqate-validate';
```
katzmo: This class isn’t used anywhere anymore. Do you want to keep it anyway?

campaignion_loqate/package/src/validator.js
katzmo: Looks very jQuery validaty now! :laughing:

torotil: Erm, I was only following your suggestion … :joy:

campaignion_loqate/campaignion_loqate.info
torotil: Yes, we need to add it globally now, because the `clientsideValidationAddCustomRules` event is only fired on the initial page load. I’ve removed the `#attach` line.

campaignion_loqate/campaignion_loqate.module
torotil: I’ve removed it.

katzmo (approved):

campaignion_loqate/package/src/input.js
torotil: I’ve tested it locally with 2 payment methods:

1. Produced a validation error in the first
2. Entered valid values in the second
3. Switched back to the first → error was still displayed
4. Switched to the second → was able to successfully submit

katzmo: Sounds perfect! :ok_hand:

katzmo (approved):

PR #302, opened 2022-05-16
Update webform_confirm_email: Simplify handling of confirmed timestamps

torotil:
With the upcoming version webform_confirm_email has a confirmed timestamp instead of just a boolean value. This can be used by campaignion_activity.

Since webform_confirm_email didn’t have timestamps for this before and campaignion_activity did, we have the more accurate data to update the timestamps in the activity tables.

The update hooks are bit complicated because webform_confirm_email is an optional dependency and we only want the hook to run if webform_confirm_email was updated recently.

### DIFF
--- a/campaignion_activity/campaignion_activity.install
+++ b/campaignion_activity/campaignion_activity.install
@@ -163,6 +163,36 @@ function campaignion_activity_uninstall() {
 }
 
 /**
+ * Implements hook_update_dependencies().
+ */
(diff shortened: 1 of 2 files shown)


torotil:
[CAMP: integration confirm email with webform export](https://trello.com/c/5MGgNeyj/1350-camp-integration-confirm-email-with-webform-export)

katzmo (approved):

PR #305, opened 2022-07-05
[fix] Admin titles: Remove double-encoding for admin titles

torotil:
`$node->title` is supposed to have the unencoded values so we need to use `$item['value']` instead of `$item['safe_value']`.


torotil:
[Internal page titles shows encoded HTML of punctuation](https://trello.com/c/P2YvOgsR/2737-internal-page-titles-shows-encoded-html-of-punctuation)

katzmo (approved):

torotil:
[CIWF: prefilling form JS](https://trello.com/c/JfSzYKzR/1263-ciwf-prefilling-form-js)

PR #307, opened 2022-07-14
[webform widget] remove X-Frame-Options header

katzmo:
Better to remove the header then setting it to an invalid value.

A `Content-Security-Policy` header should be set instead, e.g. by using the [campaignion_content_security_policy](https://github.com/moreonion/campaignion/tree/7.x-2.x/campaignion_content_security_policy) module.


katzmo:
This doesn’t seem to remove the header, Drupal just adds it again (also tried `hook_page_alter()`).

PR #313, opened 2022-12-07
Bundle Vue apps with Vite instead of Webpack

mayakokits:
Replace Webpack with Vite 3 for the Vue apps in e2t, update Vue and some dependencies. Dependencies are bundled with the apps, not used via `campaignion_vue` any more.


campaignion_vue/campaignion_vue.module
```diff
@@ -58,13 +60,20 @@ function _campaignion_vue_get_strings($language = NULL) {
 function campaignion_vue_preprocess_html(&$vars) {
   $js = drupal_add_js();
   $found = FALSE;
+  $add_dependencies = FALSE;
   foreach ($js as $path => $options) {
-    if (substr($path, -11) == '.vue.min.js') {
+    if (substr($path, -24) == 'redirects_app.vue.min.js') {
```
torotil: I’m not sure where to begin. Actually library dependencies should be handled using hook_library_info(). I don’t think a rewrite makes sense now though. I guess I’d rather go without the optimization of not loading campaignion_vue.min.js on some pages where it is actually not needed.

I guess the problem will go away anyway once the redirect_app uses vite too.

torotil (changes requested):

campaignion_email_to_target/messages_app/drupal-fixture.js
```diff
@@ -134,21 +135,19 @@ var Drupal = {
...
-    }
-    else if (index == 1) {
-      return Drupal.t(plural, args, options);
-    }
-    else {
-      args['@count[' + index + ']'] = args['@count'];
-      delete args['@count'];
-      return Drupal.t(plural.replace('@count', '@count[' + index + ']'), args, options);
+      return Drupal.t(singular, args, options)
+    } else if (index == 1) {
```
torotil: Please use [brace-style](https://eslint.org/docs/latest/rules/brace-style) stroustrup for Drupal JS.

campaignion_vue/campaignion_vue.module
torotil: Ok, talked with Florian. I think it’s better to actually migrate the redirect_app too (and remove the campaignion_vue module) instead of maintaining this in-between state.

torotil (changes requested):

campaignion_email_to_target/messages_app/drupal-fixture.js
mayakokits: Just added a commit to change this, and then I thought a bit... this is not Drupal JS, it's just a partial copy of Drupal JS that is never used in a Drupal context. It's just used by the Vue app in dev and testing mode where the `Drupal` global isn't available.
Also, the Vue app's linter stripped away all the semicolons, which might not fit the Drupal style.
So in my opinion it's cleaner to have this file linted like the rest of the Vue app. This would make my last commit obsolete.

mayakokits:
Now the `redirects_app` is bundled with Vite as well.
We still need the `campaignion_vue` module, as it provides localized strings for the Vue apps that can’t know the user's language at build time. But I cropped it down and removed all that is obsolete.

torotil (approved):

PR #314, opened 2022-12-14
Match to target: Add tokens in email_to_target.

katomat:


campaignion_email_to_target/src/MessageTemplateInstance.php
```diff
@@ -36,6 +36,7 @@ public function replaceTokens(array $target = NULL, Submission $submission = NUL
       $target['display_name'] = $target['salutation'];
     }
     $data['email-to-target'] = $target;
+    $data['target'] = $target;
```
torotil, line 38: This can actually replace `$data['email-to-target']`. There is no need to keep it around.

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -172,6 +174,9 @@ function campaignion_email_to_target_tokens($type, array $tokens, array $data =
   if ($type == 'email-to-target' && !empty($data['email-to-target'])) {
```
torotil: This should be replaced with `$data['target']` too. We entirely control how data is passed to this function so there is no need to keep the old data around.

campaignion_email_to_target/src/Tokens.php
```diff
@@ -25,12 +25,18 @@ public static function submissionTokens(array $tokens, Submission $submission) {
           $messages[] = $message;
         }
       }
-      $t = 'campaignion_email_to_target_messages_token';
-      $text = theme([$t, $t . '_' . $submission->node->nid], [
-        'messages' => $messages,
-        'submission' => $submission,
-      ]);
+      $token = [
```
torotil: These are theme hooks not tokens. This should be extended with the nid-postfixed hook too, and then be passed to `theme()` in one go.

Also we could check if any of our themes actually overrides this. If not we can just replace the old theme hook without caring for backwards compatibility.

torotil (changes requested):

campaignion_email_to_target/src/Tokens.php
katomat: This seems to be only used in [campaignion_email_to_target.module file](https://github.com/moreonion/campaignion/blob/bc2af86179e0afefa63481b537e00a5a86c4ac20/campaignion_email_to_target/campaignion_email_to_target.module#L275). I didn't find a theme overriding it - would I rather need to go through the database to check if any of the emails would contain the token?

torotil: There is no token involved in this particular part of the code. The word is only your interpretation of what the `$t` (for theme hook) meant.

katomat: Yes I got that, sorry, I missed mentioning that  this was crucial information, thank you!
My answer was more pointed toward the rest of the code not that specific bit. However I figured out that I also had renamed the template in [campaignion_email_to_target.module](https://github.com/moreonion/campaignion/blob/bc2af86179e0afefa63481b537e00a5a86c4ac20/campaignion_email_to_target/campaignion_email_to_target.module#L275) which caused the token to not take effect in the thank you email.

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -134,16 +134,18 @@ function campaignion_email_to_target_token_info() {
...
-      $type = 'email-to-target';
-      $info['types'][$type] = [
-        'name' => t('Email-to-Target dataset: %name', ['%name' => $dataset->title]),
-        'description' => t('Replace values based on the selected target(s).'),
-      ];
-      foreach ($dataset->attributes as $attribute) {
-        $info['tokens'][$type][$attribute->key] = [
-          'name' => $attribute->title,
-          'description' => $attribute->description,
+      $types = ['email-to-target', 'target'];
```
torotil: Should we keep the documetation of the `email-to-target` tokens? If so, should we add an info that it’s deprecated?

campaignion_email_to_target/campaignion_email_to_target.module
```diff
@@ -169,8 +171,8 @@ function campaignion_email_to_target_tokens($type, array $tokens, array $data =
     $submission = new Submission($node, $s);
     $replacements = Tokens::submissionTokens($tokens, $submission);
   }
-  if ($type == 'email-to-target' && !empty($data['email-to-target'])) {
-    $replacements = Tokens::messageTokens($tokens, $data['email-to-target']);
+  if ($type == ('target' || 'email-to-target') && !empty($data['target'])) {
```
torotil: This condition won’t work. It will only at the new tokens because `('target' || 'email-to-target') == 'target'`.

campaignion_email_to_target/src/Wizard/TargetStep.php
```diff
@@ -70,7 +70,7 @@ public function stepForm($form, &$form_state) {
       ],
       [
         'key' => 'salutation',
-        'description' => 'This field is how the target will be addressed in the email opening so should include any appropriate honorifics e.g. ‘Rt Hon John Smith MP’',
+        'description' => 'This field is how the target will be addressed in the message opening so should include any appropriate honorifics e.g. ‘Rt Hon John Smith MP’',
```
torotil: We will need to actually change these labels in [e2t-data](https://gitlab.more-onion.com/e2t/e2t-data) too. Otherwise they will still show up differently in the token info.

torotil (changes requested):

campaignion_email_to_target/src/Wizard/TargetStep.php
katomat: I looked through the dataset.yaml files and the descriptions differ. But I think it makes sense that it does.
e2t-data has a more generic approach, for instance:
Salutation = Spoken form of personal address; also used at start of letter.
while the Drupal code features a more help text style user centred description. Do you think they should be the same?

campaignion_email_to_target/src/Tokens.php
```diff
@@ -14,24 +14,26 @@ abstract class Tokens {
...
+        ]);
+        $replacements[$tokens[$token]] = $text;
       }
-      $themehook = 'campaignion_target_messages_token';
-      $text = theme([$themehook, $themehook . '_' . $submission->node->nid], [
-        'messages' => $messages,
-        'submission' => $submission,
-      ]);
-      $replacements[$tokens['target-messages']] = $text;
-      $replacements[$tokens['email-to-target-messages']] = $text;
```
torotil: Why was this change needed? Both version should work as far as I can tell from just reading the code. I think in this case adapting the test would have been the better option. (If the previous option made a test fail.)

torotil (approved):

PR #318, opened 2023-02-03
[email_to_target] Remove test mode for match to target type.

katomat:
This aims to have the test mode in the campaignion wizard only displayed for email-to-target content types.


katomat:
Closed in favour of 314.

PR #323, opened 2023-03-10
[campaignion_tracking] add Google Analytics 4

katzmo:


campaignion_tracking/package/src/ga4.js
```diff
@@ -0,0 +1,426 @@
...
+ * Machine readable name.
+ *
+ * Use for referencing tracker specifics like storage keys, channel names, ...
+ */
+export const name = 'ga4'
+
+/**
+ * Mapping of codes to eventNames.
+ *
+ * This also is a whitelist of which events should *also* be recognized as
```
torotil: This could easily say “list“ instead of “whitelist“ and also I think there is one more also than needed.

campaignion_tracking/package/src/ga4.js
```diff
@@ -0,0 +1,426 @@
...
+
+    this.printDebug('init')
+
+    if (typeof this.tracker === 'undefined') {
+      this.printDebug('No TrackerManager found. Doing nothing.')
+      return
+    }
+
+    if (typeof this.dataLayer === 'undefined') {
+      this.printDebug('No datalayer found. Doing nothing.')
```
torotil, line 42: Is there a return `return` missing after this statement?

campaignion_tracking/package/src/ga4.js
```diff
@@ -0,0 +1,426 @@
...
+  handle_checkoutBegin (eventName, eventData, context) {
+    this.printDebug('(handle)', eventName, eventData, context)
+    this.updateContext(context)
+    const product = eventData.product || this._context.donation.product || {}
+    const currencyCode = eventData.currencyCode || this._context.donation.currencyCode || null
+    let data = {
+      event: 'checkoutBegin',
+      ecommerce: {
+        checkout: {
+          actionField: { step: 1 }, // begin == 1
```
torotil: The comment here is funny. The one thing it says is the only thing that’s obvious from the code.

Is this something required bei GA? Is this something we define arbitrarily (and thus might need user documentation)?

campaignion_tracking/package/src/ga4.js
```diff
@@ -0,0 +1,426 @@
...
+    // Allow others to modify the data being sent to Google Analytics.
+    data = this.callChangeHook(eventName, data, this._context)
+    this.dataLayer.push(data)
+  }
+
+  /**
+   * Handle "donationSuccess".
+   *
+   * The event data needs to include an transaction id.
+   * Falls back to use a random number.
```
torotil: This comment contradicts itself: Either it is required or there is a default. It can’t be both.

campaignion_tracking/package/src/ga4.js
```diff
@@ -0,0 +1,426 @@
...
+   * Remove a (donation) product when one was added before. Thus we have a
+   * "cart" with only 1 slot for 1 donation.
+   *
+   * Event data: { product, currencyCode }
+   *
+   * @param {String} eventName the event name
+   * @param {object} eventData data of the event
+   * @param {object} context context of the event
+   */
+  handle_setDonationProduct (eventName, eventData, context) {
```
torotil, line 299: I think the logic of this function could be simplified if it would separate the product comparison from preparing the events. This also applies to the GA4 version of the next commit. So:

1. Build new product & compare it to the old, if equal then return.
2. Build events and send them.

campaignion_tracking/js/campaignion_tracking.js
```diff
@@ -8,7 +8,7 @@ Drupal.behaviors.campaignion_tracking.attach = function(context, settings) {
    * Nothing happens if the tracker is not available.
    */
 
-  function gracefulDispatch (name, data, trackingContext) {
+  function gracefulDispatch (type, name, data, trackingContext) {
```
torotil: This could just as well be defined outside the behavior which makes the behavior simpler and makes it clear it’s pure function (not a closure).

campaignion_tracking/js/campaignion_tracking.js
```diff
@@ -19,28 +19,58 @@ Drupal.behaviors.campaignion_tracking.attach = function(context, settings) {
...
+     *
+     * Events:
+     * - 'draftBegin'
+     * - 'draftContinue'
+     */
+
+    // Fire `draftBegin` after completing the first step of multi-step forms.
+    if (!settings['campaignion_tracking']['sent'].includes('draftBegin')) {
+      if (webform['last_completed_step'] === 1) {
+        settings['campaignion_tracking']['sent'].push('draftBegin');
```
torotil: Using Drupal.settings to store state is very weird it’s supposed to be only used for reading settings passed by the backend. Could `this` be used instead?

torotil (changes requested):
I’ve found a few things that I think might be worth changing. I’m not sure how far we should push this as this is mostly a c&p of existing code (with potential of improvement). :innocent: Happy to discuss what’s worth the effort and what isn’t.

campaignion_tracking/package/src/ga4.js
torotil: Also the product comparison seems like something that’s needed in several integrations. The base integration could take care of doing this and fire an event only when the product actually changes.

campaignion_tracking/package/src/ga4.js
katzmo: Looks like it.

campaignion_tracking/package/src/ga4.js
katzmo: No idea, I've assumed it means something for GTM :shrug: It’s not part of the GA4 event.

campaignion_tracking/package/src/ga4.js
katzmo: I don’t think so. The transaction id is required by google, not by this method. If no id is passed, the method will create a random id to send in the event.

campaignion_tracking/package/src/ga4.js
katzmo: I’m afraid it's a bit more complicated. As much as I'd like to reduce unnecessary "setDonationProduct" events, this would need to keep track of the previous product data for comparison. Where/How would you suggest to do that? :thinking:

campaignion_tracking/js/campaignion_tracking.js
katzmo: No, the data needs to be persistent between behavior calls. An alternative might be to use the session storage (this could also help with comparing products as discussed in a previous comment).

campaignion_tracking/package/src/ga4.js
katzmo: Not even sure the list is needed? :thinking:

campaignion_tracking/js/campaignion_tracking.js
katzmo: I think it makes it harder to find the Drupal behavior… :thinking:

campaignion_tracking/package/src/ga4.js
katzmo: Have removed it and replaced all occurrences of "whitelist".

campaignion_tracking/package/src/ga4.js
katzmo: Turns out the current version of the GA4 code didn’t even work :innocent: 

Have refactored the product comparison logic and deleting products, choosing session storage as this was already used by other bits of codes.

campaignion_tracking/js/campaignion_tracking.js
katzmo: I think sessionStorage actually works quite well!

campaignion_tracking/package/src/ga4.js
torotil: Thank you!

campaignion_tracking/package/src/ga4.js
torotil: Ok, then never mind :grimacing:

campaignion_tracking/package/src/ga4.js
torotil: > Where/How would you suggest to do that?

I’m assuming that the product data is the same for all integrations (which I didn’t check thoroughly). Then campaignion_tracking could do the comparison and each individual integration could be called only when it actually changes.

campaignion_tracking/js/campaignion_tracking.js
torotil: > I think it makes it harder to find the Drupal behavior…

It could also be below the behavior. Another method of the behaviour would also be an improvement.

campaignion_tracking/js/campaignion_tracking.js
torotil: `Drupal.behaviors.campaignion_tracking` has the same life time as `Drupal.settings`, so it would survive across AJAX calls but not across page loads.

Would the sessionStore need deletion when the donation is completed?

torotil (changes requested):

campaignion_tracking/package/src/ga4.js
torotil: Ah I see that you already did refactor that. Awesome! :clap:

campaignion_tracking/js/campaignion_tracking.js
katzmo: Hm, removing the storage data would be a good idea, otherwise the events are only triggered on the first form of a session. But when should we do that, how can the behavior know the form is complete or a different form than before? :thinking:

torotil: The former code that stored the data in settings did that automatically: The settings are kept during AJAX calls but not for new actions, because that would mean a page load.

With the sessionStore you could still check for page loads in the behavior (context=window.document) and reset the storage. That has the advantage that in theory you could also track the transaction as “cancelled“.

katzmo: Shouldn’t the tracking work without AJAX enabled on forms as well? (Up until now this would have just triggered events too often.)

katzmo: I've pushed a new commit to store the data per form id instead of removing it. This should make the tracking possible for multiple forms in the same session, with and without AJAX enabled. 

When the user comes back to a previously tracked form in the same session one could argue this continues the submission from a tracking perspective so completing certain steps is not tracked again :innocent:

campaignion_tracking/js/campaignion_tracking.js
torotil: Much better. Thanks.

campaignion_tracking/js/campaignion_tracking.js
torotil: What if someone does a one-time donation and a regular one with the same form? But yes, I’m not even sure what the behaviour should be.

campaignion_tracking/package/src/ga4.js
torotil: I don’t think it’s obvious that this comment refers to GA’s needs instead of the function’s parameters.

campaignion_tracking/js/campaignion_tracking.js
katzmo: It would sent ~~a "remove from cart" event for the one-time donation,~~ a "add to cart" event for the regular one and another "purchase" event, but not another "begin checkout" or "end checkout".

campaignion_tracking/package/src/ga4.js
katzmo: I didn't write the comment and I find it pretty clear. But you’re welcome to make a suggestion for improvement!

torotil:
```suggestion
   * GA requires a transaction ID to be included in the event. If none is
   * passed this function generates a random ID.
```

torotil: Yep. The perils  of copy&pasting code from somewhere else and putting it to review again. I’ve been on the other side of this often enough. Sorry if it’s frustrating.

campaignion_tracking/js/campaignion_tracking.js
torotil: This sounds like something that’s very hard to answer when somebody asks for it. Is there some documentation this would fit in?

katzmo: I've added a new paragraph to the README.

campaignion_tracking/package/src/ga4.js
```diff
@@ -73,6 +73,13 @@ export class GA4Tracker {
     this.donationSubscription = this.tracker.subscribe('donation', this._dispatch)
   }
 
+  /**
+   * Copy gtag function as required by GA4.
```
torotil: How about

> Generate an arguments object and pass it to the dataLayer.

This would tell us why the function is needed as well.

(github doesn’t provide me with the option to suggest an edit for this line.)

campaignion_tracking/package/src/ga4.js
```diff
@@ -451,11 +452,11 @@ export class GA4Tracker {
...
-          item_name: product?.name,
-          item_id: product?.id,
-          price: product?.price,
-          item_variant: product?.variant,
-          quantity: product?.quantity,
+          item_name: product.name,
+          item_id: product.id,
+          price: product.price,
+          item_variant: product.variant,
+          quantity: product.quantity,
```
torotil: This commit shows nicely that the generation of the product data from `eventData` and `this._context` could be done in one place and called from all 3 handlers. What do you think?

torotil (changes requested):
Thanks for the improvements! I have two more suggestions :innocent:

campaignion_tracking/package/src/ga4.js
katzmo: Have updated the comment. To me the most important info is still that it's a copy of the gtag function.

campaignion_tracking/package/src/ga4.js
katzmo: It could. Also makes more sense to store the formatted data than the raw one. Have pushed a new commit.

katomat (approved):

torotil (approved):

PR #325, opened 2023-03-31
Draft: Mailchimp compressed responses

alex-mo:
Mailchimp can send responses compressed (with `gzip`).

`ApiError` needs to deal with it to properly recognize error cases. If it fails recognizing them, working the queue will eventually block due to not cleaned up, persistently erroring queue items.


torotil:
Closing in favor of moreonion/little_helpers#27

PR #327, opened 2023-05-25
Add explanation of when in Impact Stack the GA4 events happen

edjw:
This is to make this a bit more understandable for support, client-care, clients

### DIFF
--- a/campaignion_tracking/campaignion_tracking_ga4/README.md
+++ b/campaignion_tracking/campaignion_tracking_ga4/README.md
@@ -15,7 +15,7 @@ included in a different way, e.g. via the profile.
 
 ```json
 {
-  "event": "begin_action",      // triggered by draftBegin
+  "event": "begin_action",      // triggered by draftBegin (fired once after completing the first form step)
   "params": {
     "nid": "135",               // node id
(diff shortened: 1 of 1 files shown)


katomat (approved):
Thank you, yes I also think it is useful to have this more detailed in the readme!

katzmo:
Actually, the information when each custom event is triggered is [in the generic README](https://github.com/moreonion/campaignion/tree/7.x-2.x/campaignion_tracking/package#events). The service specific implementation documents which implementation specific event is triggered by what generic custom event. I see how it can be helpful to document both in 1 place, but it also makes it harder to keep the documentation up to date (as a change to generic events needs a change in the sub-modules too, all of them!).

@edjw Could you please update the comments for GTM as well to have some consistency? And please fix "remove_from_cart" → `setDonationAmount` is always triggered the same way: choosing both a donation amount and interval. When this happens for the 1st time, there is nothing to remove, but any time the amount or interval changes, the old values are removed and the new values are added. With the default form settings it is not possible to deselect them.

edjw:
I reverted this change but I don't plan on coming back to this

PR #334, opened 2024-07-18
Implement node cloning

torotil:


campaignion_manage/src/ContentPage.php
```diff
@@ -47,6 +48,9 @@ public function __construct(Content $query) {
       $ops['publish'] = new ContentPublish();
       $ops['unpublish'] = new ContentUnpublish();
     }
+    if (module_exists('campaignion_manage_clone')) {
```
katzmo, line 51: Wouldn’t `use Drupal\campaignion_manage_clone\ContentClone;` throw an error if the module didn’t exist?

torotil: I don’t think so. The `use` statements work a bit different than imports in Python. They only clarify define the meaning of the name for the rest of the script.

Use of an undefined class

katzmo: Ok, fine then :) I also managed to find a comment on the documentation page explaining this (there is nothing in the documentation): https://www.php.net/manual/en/language.namespaces.importing.php#121045

katzmo (approved):

katzmo (approved):

PR #340, opened 2024-11-14
fix(widget): Replace DOMSubTreeModified with MutationObserver

torotil:
DOMSubTreeModified has been deprecated since years and  stopped working in recent Chrome builds.

### DIFF
--- a/campaignion_webform_widget/js/widget.js
+++ b/campaignion_webform_widget/js/widget.js
@@ -58,9 +58,10 @@ window.addEventListener("DOMContentLoaded", function() {
 
 window.onload = function() {
   messageParent(false);
-  window.addEventListener("DOMSubtreeModified", function() {
+  // Notify the parent window of changed content after AJAX requests.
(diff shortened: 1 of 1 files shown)


campaignion_webform_widget/js/widget.js
```diff
@@ -56,11 +56,12 @@ window.addEventListener("DOMContentLoaded", function() {
   messageParent(false);
 });
 
+// Update iframe size after AJAX requests.
```
katzmo: I find the comment a little misleading. Sure, it would be the desired effect, but the function only updates the iframe height info where the embedding page can pick it up or ignore…

torotil: I’ve changed the comment and moved it to a better place. Do you think it’s better now?

katzmo: Perfect, thanks!

katzmo (approved):

PR #342, opened 2024-12-19
fix: Avoid v-html for highlighted text

torotil:
NB: This removes the hereto unused parameter `template` from the DestinationField components.


campaignion_wizard/redirects_app/src/components/HighlightedText.vue
```diff
@@ -0,0 +1,37 @@
...
+</docs>
+
+<template>
+ <span>
+ <template v-for="part in getParts(text, search)">
+  <strong v-if="part.highlight">{{ part.text }}</strong>
+  <template v-else>{{ part.text }}</template>
+ </template>
+ </span>
+</template>
```
katzmo, line 13: The indentation seems off to me, I would expect
```html
<template>
  <span>
    <template v-for="part in getParts(text, search)">
      <strong v-if="part.highlight">{{ part.text }}</strong>
      <template v-else>{{ part.text }}</template>
    </template>
  </span>
</template>
```
(same for e2t)

katzmo (approved):
Thanks, this change is not only CSP compliant, it also makes the code a lot easier to understand!

katzmo (commented):
Also, the DestinationField test still looks for `c.computed.templateComp` (that's why the unit tests fail).

katzmo (approved):
Thanks!

PR #344, opened 2025-02-12
gtm/CSP: Allow GTM in script-src

mojulia:


torotil (approved):

torotil:
This shouldn’t be needed with moreonion/drupal-d7csp#4

PR #350, opened 2025-04-11
Google CSP directives

mojulia:


torotil:
Even if it works like this as well this should be one array entry per host. You can use something like:

```php
$hosts['connect-src'] = array_merge($hosts['connect_src'], ['more', 'stuff', 'here']);
```

If `img-src` and `connect-src` need almost the same entries I would make that more obvious by assigning common hosts into a variable that is then merged into both.

Perhaps it would be even better to just copy & paste the output of https://www.google.com/supported_domains into a string constant and then reformat in the hook. Something like:

```php
// This is copy & pasted from https://www.google.com/supported_domains
define('GA4_DOMAINS', 'copy & paste here');

function campaignion_tracking_ga4_d7csp_hosts() {
  // …
  $domain_hosts = array_map(function ($domain) {
    return "https://*$domain";
  }, explode(' ', GA4_DOMAINS));
  // …
}
```

mojulia:
Oh that was very fast! Yes yes, you're absolutely right, this needs some work - thanks for the suggestions, I will look into it more closely next week :)

torotil (approved):

